PR #110, opened 2015-01-06
Extending the meta-tag for news_keywords with default keywords and terms...

### LINKED ISSUE #107: Add default keywords to meta news_keywords
The news-sitemap.xml shows both the keywords set in the post and the default keywords.
xml_news_sitemap

The meta tag only shows the keywords set in the post.
view-source_theme_dev_news-2014-12-16-2_

Please add the default keywords as well in the latter case.

### LINKED ISSUE #108: Add tags to meta news_keywords
The news-sitemap.xml shows the keywords set in the post, the default keywords and tags.
xml_news_sitemap

The meta tag only shows the keywords set in the post.
view-source_theme_dev_news-2014-12-16-2_

Please add the tags as well in the latter case.

(Issue is similar to #107)

andizer:
... from the news item. 

[Finishes #84657778] [Finishes #84657502] Fixes #107, fixes #108

### DIFF
--- a/classes/class-meta-keywords.php
+++ b/classes/class-meta-keywords.php
@@ -24,6 +24,9 @@ class WPSEO_News_Meta_Keywords {
 		// Trim each keyword
 		$keywords = array_map( array(self, 'parse_keyword'), $keywords );
 
+		// Make the list of keywords unique
+		$keywords = array_unique( $keywords );
+
 		// If keywords should be returned as string, implode a comma between each keyword
 		if ( $as_string ) {
 			$keywords = trim( implode( ', ', $keywords ), ', ' );
--- a/classes/class-meta-keywords.php
+++ b/classes/class-meta-keywords.php
@@ -1,0 +1,88 @@
+<?php
+
+class WPSEO_News_Meta_Keywords {
+
+	/**
+	 * Getting the keywords for given $item_id
+	 *
+	 * @param integer $item_id
+	 * @param bool    $as_string
+	 *
+	 * @return string
+	 */
+	public static function list_keywords( $item_id, $as_string = true ) {
+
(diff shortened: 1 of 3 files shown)


classes/class-meta-keywords.php
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	/**
+	 * This method will lowercase the whole keyword and trim spaces before and after it.
+	 *
+	 * @param string $keyword
+	 *
+	 * @return string
+	 */
+	private static function parse_keyword($keyword) {
```
omarreiss: I'd prefer naming this function `sanitize_keyword`. I'd use the word "parse" when I am really breaking down something into different bits and pieces. That's not the case here, you're just stripping whitespaces and lowercasing, basically "sanitizing" the strings for consistent representation.

classes/class-sitemap.php
```diff
@@ -88,7 +88,7 @@ public function build() {
 
 				}
 
-				$keywords      = $this->get_item_keywords( $item->ID );
+				$keywords      = WPSEO_News_Meta_Keywords::list_keywords( $item->ID );
```
omarreiss: I'd prefer this to be the following:

```
new WPSEO_News_Meta_Keywords( $item->ID )->keywords;
```

That way `WPSEO_News_Meta_Keywords` would act more like a true value object which I think it is in this case. Also it would help clean up the `WPSEO_News_Meta_Keywords` class since we'd no longer have to boundlessly pass around the `$keywords` but we can simply set a property on the value object. That'd definitely be better design. The question is, how much would we suffer from that performancewise? @jdevalk please comment on this.

omarreiss:
@andizer Nice work!

PR #118, opened 2015-02-19
Bugfix: changed attachment URL generation

mjaschen:
$attachment->guid is not necessarily the actual image URL and should not
used to extract an attachment image URL.

I changed the code to use the function wp_get_attachment_url() which is
the documented way to get an URL for an attachment.

References:

http://pods.io/2013/07/17/dont-use-the-guid-field-ever-ever-ever/
http://codex.wordpress.org/Function_Reference/wp_get_attachment_url


andizer:
Looks great, thanks :+1:

PR #120, opened 2015-02-23
Change the hook to change the sitemap URL [Delivers #86108310]

### LINKED ISSUE #82: wpseo_news_sitemap_url doesn't work properly
When using ```wpseo_news_sitemap_url``` the link to the sitemap in de admin is changed, but not the location of the sitemap itself. 

For example:
When setting the following in functions.php
```php
function tacoverdo_change_news_sitemap() {
	return home_url() . "/test_news_sitemap.xml";
}
add_filter('wpseo_news_sitemap_url', 'tacoverdo_change_news_sitemap');
```
Link found in the admin (button): http://example.com/test_news_sitemap.xml
Actual location of the sitemap: http://example.com/news_sitemap.xml

andizer:
Fixes #82 

I changed the way the hook for changing the news sitemap URL. This solution makes the sitemap accessible, instead of only directing it to that URL. 

@jdevalk Can you give me feedback if my approach for this problem is correct and my solution is acceptable?


omarreiss:
CR :heavy_check_mark:

PR #126, opened 2015-03-17
Stories/cg/check cat array before loop [finishes #90493004] [fixes #125]

### LINKED ISSUE #125: Check if $cats is an array before looping through it
The following error was reported by a customer:
Warning: Invalid argument supplied for foreach() in /var/www/web482/web/wp-content/plugins/wpseo-news/classes/class-sitemap.php on line 79

In classes/class-sitemap.php we do not check if $cats is an array before looping through it. $cats returns an array, false or a wp_error. We should check if it's an array before looping through it.

CarolineGeven:
Check if $cats is an array before looping through it, as $cats = [get_the_terms](http://codex.wordpress.org/Function_Reference/get_the_terms), which returns an array, false or a wp_error. [fixes #125]

### DIFF
--- a/classes/class-sitemap.php
+++ b/classes/class-sitemap.php
@@ -270,7 +270,7 @@ class WPSEO_News_Sitemap_Item {
 		$cats    = get_the_terms( $this->item->ID, 'category' );
 		$exclude = 0;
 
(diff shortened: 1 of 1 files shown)


andizer:
CR Done :+1:

PR #133, opened 2015-04-24
Fix the stylesheet when sitemap is served from cache

### LINKED ISSUE #132: NaN in news-sitemap - causes by loading wrong XSL
After upgrading from News SEO 2.2.1 to 2.2.3 the following errors show in the news sitemap, possibly causing problems in Google News:
xml_sitemap

Possible cause is loading the wrong xsl-file, 
`<?xml-stylesheet type="text/xsl" href="//clean.dev/main-sitemap.xsl"?>`
instead of 
`<?xml-stylesheet type="text/xsl" href="//clean.dev/wp-content/plugins/wpseo-news/assets/xml-news-sitemap.xsl"?>`

WordPress version 4.2
WordPress SEO version 2.1.1

andizer:
Fixes #132


jdevalk:
CR: :+1:

PR #135, opened 2015-04-28
Remove htmlspecialchars from output where cdata is used

### LINKED ISSUE #131: Escape HTML for content wrapped in CDATA tags?
We received the following report from a customer.
> I had a post with the title "Wells Fargo & Co. Blog Post"
> Which was being reinterpreted inside the XML to "Wells Fargo &amp; Co. Blog Post" automatically.
> Example: 		`<news:title><![CDATA[Wells Fargo &amp; Co. Blog Post]]></news:title>`
> Within CDATA the transformation to &amp; is not necessary and may actually create issues since there is no processing within CDATA
> So, shouldn't it be 		`<news:title><![CDATA[Wells Fargo & Co. Blog Post]]></news:title>`

@jdevalk / @omarreiss: Should we change the way this is currently handled?

Petervw:
Fixes #131

### DIFF
--- a/classes/class-sitemap.php
+++ b/classes/class-sitemap.php
@@ -319,18 +319,18 @@ class WPSEO_News_Sitemap_Item {
 		$this->build_publication_tag();
 
 		if ( ! empty( $genre ) ) {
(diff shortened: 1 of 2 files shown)


jdevalk:
CR: :+1:

PR #149, opened 2015-07-31
Loading library from another CDN.

### LINKED ISSUE #148: jquery.tablesorter.min.js not available
News SEO tries to load `jquery.tablesorter.min.js` from http://cdn.yoast.com/js/jquery.tablesorter.min.js, however that file no longer exists, causing a JavaScript warning on every installation of News SEO. 

This results in two errors.
developer_tools_-_http___clean_dev_news-sitemap_xml

andizer:
Fixes #148

@omarreiss The libraries are only loaded for zebrastriping. Is such a layout-specific thing desirable for a sitemap that mainly be loaded by machines (google and so?).

I can also imagine that these kind of files are being loaded from a CDN because of traffic, but it hard to rely on a CDN just because there's some need for fancyness.  Is shipping the libraries better than using CDN? What's your opinion.


omarreiss:
CR :+1: according to @andizer and @omarreiss.

omarreiss:
For Acceptance: Please test the if the news sitemap zebra striping and tablesorting works.

PR #156, opened 2015-10-28
Only add the tab hooks when the post type is supported.

andizer:


CarolineGeven:
When testing the branch with yoast seo trunk, the following fatal occurs:
schermafbeelding 2015-10-29 om 11 41 46

PR #178, opened 2016-01-07
Converted the plugin to now work via Composer

jcomack:
Converted the plugin to now work via Composer and moved the main class to the classes folder.
Also added the old autoloading method for the sake of backwards compatibility.


atimmer:
Code looks good, but the build should pass.

wpseo-news.php
```diff
@@ -30,349 +30,15 @@
...
-			return wpseo_xml_sitemaps_base_url( $sitemap_name . '-sitemap.xml' );
-		}
-
-		return $sitemap_name;
+if ( file_exists( dirname( __FILE__ ) . '/vendor/autoload_52.php' ) ) {
+	require dirname( __FILE__ ) . '/vendor/autoload_52.php';
+}
 
-	}
+if ( ! file_exists( dirname( __FILE__ ) . '/vendor/autoload_52.php' ) ) {
```
omarreiss: Please remove these lines and remove the legacy autoloader. If we let composer do the autoloading, then we let composer do the autoloading. That was one of the main reasons for composerifying anyway.

omarreiss:
@jcomack CR done. One issue found.

andizer:
@omarreiss There is something wrong with the beta script, it won't call composer install.

andizer:
@omarreiss  I've just fixed it in the beta script.

andizer:
Acceptance done :+1:

PR #180, opened 2016-01-12
Remove query call on all pages except post edit

### LINKED ISSUE #145: Unnecessary WP_Query calls made in Metabox Instantiation
WPSEO_News_Meta_Box::add_meta_fields_to_wpseo_meta is called on queries to view posts. Unless I'm missing something, this doesn't seem necessary as the meta boxes will only be visible on the edit.php/edit-new.php. This isn't a huge deal as most of the time meta box creation is cheap.

But wpseo-news does a WP_Query as part of standouts_used(). As going to the database is (relatively) time consuming, it would be good to avoid doing this. I've added a simple check to see whether it's a page that actually displays the meta boxes, happy to do a PR if you guys are interested.

CarolineGeven:
Fixes #145 

When the meta box got registered, it did a query for the description.
This is unnecessary on all pages except post edit page.

One concern with this solution is, if there would be default values in
the metabox registration, the defaults wouldn't be taken into account
when getting the values. But because News SEO doesn't have any defaults,
this is not an issue.

### DIFF
--- a/classes/class-meta-box.php
+++ b/classes/class-meta-box.php
@@ -10,11 +10,16 @@ class WPSEO_News_Meta_Box extends WPSEO_Metabox {
 	private $max_standouts = 7;
 
 	public function __construct() {
+		global $pagenow;
+
 		$this->options = WPSEO_News::get_options();
 
 		add_filter( 'wpseo_save_metaboxes', array( $this, 'save' ), 10, 1 );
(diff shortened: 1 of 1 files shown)


omarreiss:
CR :+1:

andizer:
Acceptance done :+1:

PR #189, opened 2016-02-13
Remove the autoloader, composer is now used

atimmer:
We don't need to ship this if it is no longer used.


atimmer:
Build is failing because `release/3.1` is also failing. The reason it's failing is because of code style, which is fixed in `trunk`.

PR #193, opened 2016-02-25
wpseo_admin_pages is no longer being used in Yoast SEO. So it can be removed

diedexx:
related to https://github.com/Yoast/wordpress-seo-local/issues/458#issuecomment-188829155
Just like in Local SEO, this filter is no longer relevant.


jcomack:
CR done

terw-dan:
Since there is an error in news with a 404 on admin-page.min.js this issue is blocked.

diedexx:
@terw-dan: The 404 error you get is fixed in #198

atimmer:
👍🏻

PR #230, opened 2016-07-05
Use better red.

### LINKED ISSUE #229: Update the reds (colors!)
Some of the red shades are `#f00` and don't have a sufficient color contrast ratio with the white background. As already done in other parts, see for example https://github.com/Yoast/wordpress-seo/issues/4058, we should standardise on the new WordPress red `#dc3232` for white backgrounds.

screen shot 2016-07-05 at 16 11 27

About this specific error message, maybe also toggle a CSS class instead of using inline styles.

afercia:
Fixes #229.
Also, some coding standards.


classes/class-wpseo-news.php
```diff
@@ -243,7 +243,7 @@ public function enqueue_edit_post() {
 	 * Print the edit post CSS
 	 */
 	public function edit_post_css() {
-		echo "<style type='text/css'>.wpseo-news-input-error{border:1px solid #ff0000 !important;}</style>" . PHP_EOL;
+		echo "<style type='text/css'>.wpseo-news-input-error{border:1px solid #dc3232 !important;}.wpseo-news-error-message{color:#dc3232;font-weight:bold;}</style>" . PHP_EOL;
```
terw-dan: Can't we do without the !important here?

terw-dan:
CR done, only 1 small question @afercia

afercia:
Alternate approach. Was thinking we could use the native WordPress CSS classes `form-invalid` and `error-message` so we could also remove the inline stylesheet. However, these WordPress CSS classes might be removed in the future (?) and expose us to maintenance issues. Thoughts?
Otherwise, we can simply keep the inline style but we should use a selector with a higher specificity or `!important`.

afercia:
Forgot to say that the WordPress `error-message` uses a `#f00` color but that should be fixed upstream and there's a ticket for that: https://core.trac.wordpress.org/ticket/35622

atimmer:
CR Done 👍🏻

atimmer:
Acceptance done 👍🏻, @afercia I got some merge conflicts, can you fix them so I can merge this?

afercia:
Merged conflicts cc @atimmer

PR #232, opened 2016-07-13
Add hook also for the sitemap, because it has to be loaded

### LINKED ISSUE #216: Individual Post/Page Genre not override Default
**What Was expected**
When an individual genre was selected on a post or page it would override the default genre. 

**What happened Instead**
The post/page genre did not override the default genre and did not appear in the News SEO sitemaps' meta-data.

**Steps to Reproduce**
1.Select a default genre. In this example, "Press-Release" has been selected.

screen shot 2016-05-25 at 3 05 21 pm

2.Select a post and make the post's genre something else. Update the post. In this example the post's name is "Video Custom Field Test" and the genre is "op-ed".

screen shot 2016-05-25 at 3 07 40 pm

3.Check sitemap for "Op-Ed", it does not appear

screen shot 2016-05-25 at 3 07 01 pm

4.Check meta data in the source-code of post. Notice that "press-release" has not been overridden

screen shot 2016-05-25 at 3 15 37 pm

**Notes:** 
Tested with all genre types on a post. They all result in the same outcome. 
Tested on a new post. It resulted in the same outcome.
Tested on pages. It resulted in the same outcome.

**Tech Specs:**
WordPress: 4.5.2
Yoast SEO Premium: 3.2.5
News SEO: 3.2

andizer:
The hook to add metafields was only edit when being on a post add/edit page. This resulted in errors when opening sitemap, because a value couldn't be found. In order to fix this I've added news-sitemap.xml also.

Fixes #216

### DIFF
--- a/classes/class-meta-box.php
+++ b/classes/class-meta-box.php
@@ -17,7 +17,7 @@ class WPSEO_News_Meta_Box extends WPSEO_Metabox {
 		add_filter( 'wpseo_save_metaboxes', array( $this, 'save' ), 10, 1 );
 		add_action( 'add_meta_boxes', array( $this, 'add_tab_hooks' ) );
 
-		if ( $pagenow === 'post.php' || $pagenow === 'post-new.php' ) {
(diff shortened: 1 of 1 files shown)


terw-dan:
CR done :+1:

boblinthorst:
Acceptance: ok 👍

PR #233, opened 2016-07-14
Publishdate w3c format

### LINKED ISSUE #112: Save the publication date in sitemap always in W3C Format
Instead of saving the local publication date in the sitemap, save the UTC standard. This way the publication date is a standard and is always sent to Google the correct way.

andizer:
The publishdate should be in W3C format when the timezone is set in UTC-format. When the timezone is set in UTC format, give the user a notice.

Fixes #112


terw-dan:
CR done :+1:

boblinthorst:
Acceptance: ok 👍

PR #257, opened 2016-10-31
Add wpseo_news_set_item filter to set_item function

maxinacube:
## Summary

Add `wpseo_news_set_item` filter to `set_item` function

* This filter allows modification of the feed item

## Test instructions

* Include `add_filter( 'wpseo_news_set_item', 'wpseo_news_set_item_textdomain' )` in theme's functions.php
* Include `wpseo_news_set_item_textdomain` function in theme's functions.php

```
function wpseo_news_set_item_textdomain( $item ) {
  if ( /* condition */ ) {
    $item['creator'] = 'condition';
  }

  return $item;
}
```

```
function wpseo_news_set_item_textdomain( $item ) {
  if ( $author_byline = get_post_meta( get_the_ID(), '_author_byline', true ) ) {
    $item['creator'] = $author_byline;
  }

  return $item;
}
```

## Fixes

Allows theme's with an author byline to filter the author's name in the `editors-pick.rss`

### DIFF
--- a/classes/class-sitemap-editors-pick.php
+++ b/classes/class-sitemap-editors-pick.php
@@ -117,13 +117,15 @@ class WPSEO_News_Sitemap_Editors_Pick {
 	 * Add a single item to $this->items
 	 */
 	private function set_item() {
-		$this->items[] = array(
+		$wpseo_news_item = apply_filters( 'wpseo_news_set_item', array(
 			'title'        => get_the_title(),
 			'link'         => get_permalink(),
 			'description'  => get_the_excerpt(),
 			'creator'      => get_the_author_meta( 'display_name' ),
 			'published_on' => get_the_date( DATE_RFC822 ),
-		);
+		) );
+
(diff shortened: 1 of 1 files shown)


moorscode:
Hi @maxinacube 

Thank you for taking the time to create this patch!

As I'm looking through your description and solution, I notice that the only thing that would be expected to change.

When passing an entire array through a filter, doing sanity checks on the filtered output is something we would need to do to make sure everything is still as expected.

I see you need to show a different author, but when this happens it should be done for all places where this would be needed.

I would think that hooking a filter onto the `get_the_author_display_name` would provide the functionality to override the display name for the post on all places where it would be used.

This can be implemented with some `is_admin` and `get_the_ID()` function calls to determine the context the filter is called in.

moorscode:
I'm closing this PR as the functionality can be achieved in a different way which covers the described problem in a better way.

Please feel free to create an issue if the mentioned solution is not solving the problem.

PR #268, opened 2016-12-15
Changed 'seven' to '7' in message about standout tag

### LINKED ISSUE #186: Inconsistent numbering: 7 per seven days
In the news tab in the Yoast SEO metabox under a post...
screen shot 2016-02-04 at 15 34 31
I would suggest changing "seven" to "7" to increase consistency.

Using wpseo-news trunk (3.0)

CarolineGeven:
After discussing this with @IreneStr, we decided to write '7 days'
instead of 'seven days'. This way there is more consistency.

Fixes #186


terw-dan:
CR 👍

PR #270, opened 2016-12-23
Invalides only on supported Post Types

moorscode:
## Summary

Only invalidate sitemap cache when an object in a Post Type that is marked as News is updated.

*

## Test instructions

* Before this patch 'Pages' triggered News sitemap invalidations, with the patch applied they should not (unless manually configured in the News SEO options)

Fixes https://github.com/Yoast/wpseo-news/issues/271


terw-dan:
Acceptance: 
When posting a page, it still seems to invalidate the sitemap.

atimmer:
I have tested this and it seems to work like it should.

PR #288, opened 2017-03-14
solve dep conflict with yoast/wordpress-seo 4.4

raufeisen:
Can only install one of: yoast/i18n-module[2.0.1, 1.2.1]

## Summary

solve i18n-module dependency conflict with yoast/wordpress-seo 4.4

*

## Test instructions

*

Fixes #

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "xrstf/composer-php52": "^1.0.20",
         "yoast/wp-helpscout": "^1.0",
-        "yoast/i18n-module": "^1.2.1"
+        "yoast/i18n-module": "^2.0.1"
     },
     "require-dev": {
         "yoast/yoastcs": "0.3",


atimmer:
The composer.lock file also needs to be updated to reflect this change.

moorscode:
This change also requires the classname to be used to be postfixed with `_v2`

moorscode:
@raufeisen thank you for making this PR.
If you could fix the two mentioned things, this can be merged into the plugin.

To update the composer.lock file, you need to do a `composer require yoast/i18n-module:^2.0` and commit both the `composer.lock` as the `composer.json` files.

To postfix the classname, you need to find where the `yoast_i18n` class is used and change it to the classname used in the v2.0 release of the i18n module.

Many thanks in advance!

jcomack:
Closing this in favor of #315

PR #293, opened 2017-05-05
Add filter for exclude_item_terms to allow custom exclusion

### LINKED ISSUE #292: [Feature request] - Add post tags to include
We've recently received a feature from one of our users. The user would like to organize the new posts by using the tags instead of the categories or post types. The user is wondering if we'd be willing to modify `exclude_item_terms` to use a filter so the user could hook into it more easily.

The user can also send a patch if that's useful.

Alternatively, can we have the option *Post tags to include/exclude* in the News SEO settings?

kitwestneat:
In order to exclude/include posts by custom criteria (for example, tags), this
patch adds a filter named 'wpseo_news_sitemap_exclude_item' that can be used to
control exclusion.
Fixes #292

### DIFF
--- a/classes/class-sitemap.php
+++ b/classes/class-sitemap.php
@@ -422,7 +422,8 @@ class WPSEO_News_Sitemap_Item {
 			return true;
 		}
 
-		if ( 'post' === $this->item->post_type && $this->exclude_item_terms() ) {
+		$exclude_item = $this->exclude_item_terms() || apply_filters( 'wpseo_news_sitemap_exclude_item', false, $this->item->ID );
+		if ( 'post' === $this->item->post_type && $exclude_item ) {
 			return true;
 		}
 


kitwestneat:
A more general purpose solution to #292 could be converting get_items to wp_query and filtering the wp_query parameters array, which would have the added benefit of excluding the items in the database as opposed to in a PHP loop. I wasn't sure if there was a reason wp_query wasn't used already though.

Pcosta88:
related: https://github.com/Yoast/wpseo-news/issues/300

classes/class-sitemap.php
```diff
@@ -422,7 +422,8 @@ private function skip_build_item() {
 			return true;
 		}
 
-		if ( 'post' === $this->item->post_type && $this->exclude_item_terms() ) {
+		$exclude_item = $this->exclude_item_terms() || apply_filters( 'wpseo_news_sitemap_exclude_item', false, $this->item->ID );
```
moorscode: If we are introducing an extra method of excluding items, it should not be merged into the other condition.

Doing the filter call before `excluding_item_terms` would be preferable, as this would not be implemented by default.

```
if ( true === apply_filters( 'wpseo_news_sitemap_exclude_item', false, $this->item->ID ) ) {
    return true;
}
```

moorscode: Could you add a comment explaining the filter and the parameters you are supplying for it?

(For an example please see: https://github.com/Yoast/wpseo-news/blob/c551d77dcc8d56a7424382be9df358d2dccbbe64/classes/class-wpseo-news.php#L14)

moorscode:
Hi @kitwestneat 

Thank you for taking the time to create this patch!

The reason why we are not using `wp_query` is that it's slow to begin with, and doing the custom date query would be hard to get in there.

I have left a review of your code, if you could update your patch to match the comments we can test it and merge it after everything checks out as expected.

If you could also merge trunk into your branch, the Travis build should be able to pass again as well.

Many thanks in advance!

PR #310, opened 2017-07-26
Deprecate method for getting the sitemap name.

andizer:
## Summary

* Deprecated `WPSEO_News::get_sitemap_name` because we have `WPSEO_News_Sitemap::get_sitemap_name` that does the same magic.

* This method  `WPSEO_News::get_sitemap_name` was calling a deprecated function. I was replacing the deprecated call by the one that replaces it, but found out there is a similar method (`WPSEO_News_Sitemap::get_sitemap_name`) that does the same. Because of there are no calls to `WPSEO_News::get_sitemap_name` I deprecated it.

## Test instructions

* Check if everything stil works.

### DIFF
--- a/classes/class-wpseo-news.php
+++ b/classes/class-wpseo-news.php
@@ -347,20 +347,16 @@ class WPSEO_News {
 	/**
 	 * Getting the name for the sitemap, if $full_path is true, it will return the full path
 	 *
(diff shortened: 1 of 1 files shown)


boblinthorst:
CR and acceptance 👍

PR #335, opened 2018-01-23
Load xsl stylesheet inline

### LINKED ISSUE #334: Make XSL stylesheet static
See https://github.com/Yoast/wordpress-seo/issues/8521 for reference

andizer:
## Summary

* Load the XSL stylesheet from a static file. This resulted in removing the tablesorter dependency.

## Test instructions

* First of all: remove all transients, to make sure no cache has been loaded.
* Checkout trunk
* Open the news sitemap (you might have to add a post to have 'news') and see tablesorting is working
* Remove all transients again
* Checkout this branch and see the table sorting not working.

Whenever the home URL and site URL are set to different values (local.wordpress.test/local.wordpress-home.test) the XSL needs to be loaded via the previously implemented method.
Otherwise the domain, protocol or port will mismatch and the XSL cannot be loaded.
* Configure home URL to be different from site URL and see the XSL being loaded from the home URL base (`{home_url}/news-sitemap.xsl`)

Fixes #334

### DIFF
--- a/assets/xml-news-sitemap-xsl.php
+++ b/assets/xml-news-sitemap-xsl.php
@@ -14,14 +14,6 @@
 			<head>
 				<title>XML News Sitemap</title>
 				<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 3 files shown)


andizer:
CR & Acceptance done 👍

PR #360, opened 2018-02-11
CS/SQL: refactor query using ` IN ()`

jrfnl:
## Summary
* Security hardening

Queries with a variable number of replacements can be prepared as well.

The placeholders are dynamically created using an `implode( ',', array_fill( 0, $count, '%s' ) )` and the complete set of replacements need to be passed as an array in the second parameter of `prepare()`.

Includes:
* Prepare late, i.e. wrap the call to `prepare()` within the query call.
* Remove a `private` method which previously did some work for the `IN` clause.

## Test instructions
* Verify that the sitemap still displays all the items expected.
* With a debug tool - check that the SQL query created and run is correct.

### DIFF
--- a/classes/class-sitemap.php
+++ b/classes/class-sitemap.php
@@ -196,28 +196,29 @@ class WPSEO_News_Sitemap {
 	private function get_items( $limit = 1000 ) {
 		global $wpdb;
 
-		$limit = max( 1, min( 1000, $limit ) );
-
-		$post_types = $this->get_post_types();
+		// Get supported post types.
+		$post_types = WPSEO_News::get_included_post_types();
 		if ( empty( $post_types ) ) {
 			return array();
 		}
 
+		$replacements   = $post_types;
+		$replacements[] = max( 1, min( 1000, $limit ) );
+
 		// Get posts for the last two days only, credit to Alex Moss for this code.
-		// @codingStandardsIgnoreStart
-		$sql_query = "
-			 SELECT ID, post_content, post_name, post_author, post_parent, post_date_gmt, post_date, post_date_gmt, post_title, post_type
-			 FROM {$wpdb->posts}
-			 WHERE post_status=%s
-			 AND (DATEDIFF(CURDATE(), post_date_gmt)<=2)
-			 AND post_type IN ({$post_types})
-			 ORDER BY post_date_gmt DESC
-			 LIMIT 0, {$limit}
-		 ";
-
-		$items = $wpdb->get_results( $wpdb->prepare( $sql_query, 'publish' ) );
-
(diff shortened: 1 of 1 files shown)


classes/class-sitemap.php
```diff
@@ -196,28 +196,29 @@ private function get_stylesheet_line() {
...
 
-		// @codingStandardsIgnoreEnd
+		// Get posts for the last two days only, credit to Alex Moss for this code.
+		$items = $wpdb->get_results(
+			$wpdb->prepare(
+				"SELECT ID, post_content, post_name, post_author, post_parent, post_date_gmt, post_date, post_date_gmt, post_title, post_type
+				FROM {$wpdb->posts}
+				WHERE post_status='publish'
+					AND (DATEDIFF(CURDATE(), post_date_gmt)<=2)
+					AND post_type IN (" . implode( ',', array_fill( 0, count( $post_types ), '%s' ) ) . ')
```
jdevalk: I think this whole `implode` etc. is kinda hard to follow when reading, can't we do this outside of the `prepare`?

classes/class-sitemap.php
```diff
@@ -196,28 +196,29 @@ private function get_stylesheet_line() {
...
-			 WHERE post_status=%s
-			 AND (DATEDIFF(CURDATE(), post_date_gmt)<=2)
-			 AND post_type IN ({$post_types})
-			 ORDER BY post_date_gmt DESC
-			 LIMIT 0, {$limit}
-		 ";
-
-		$items = $wpdb->get_results( $wpdb->prepare( $sql_query, 'publish' ) );
+		$replacements   = $post_types;
+		$replacements[] = max( 1, min( 1000, $limit ) );
```
jdevalk: What is this for? A comment on what this means would be useful for me :)

classes/class-sitemap.php
jrfnl: > can't we do this outside of the prepare?

@jdevalk Not really. If you would, you would still need to pass a variable into the query which will get flagged as needing preparation, but can't use prepare, as the variable being passed in is the string with the placeholders to do the replacements with....
Creating the query with a `sprintf()` before doing the prepare won't work either as then we'd get tripped up by the placeholder in the `limit` clause.

classes/class-sitemap.php
jrfnl: @joost Agreed, except I only moved that code. I didn't write nor change it.

The way I read it:
* `$limit` is an arbitrary number;
* the target number should be between 1 and 1000;
* the snippet makes sure between those three numbers - `1`, `1000`, `$limit` - that the highest number is selected, but no higher than 1000.

PR #361, opened 2018-02-11
CS/QA: consistently use `WPSEO_NEWS_FILE` constant

jrfnl:
## Summary
* Settings page link was missing on the plugins page.

The `WPSEO_NEWS_FILE` constant is created for a reason: always having easy access to the filename of the "main" plugin file as registered with WP.

For the activation hooks, this is just a case of consistently using the constant.

For the second commit change - using `__FILE__` should cause WP not to recognize the plugin correctly and therefore not to display the settings link on the plugins page.
I haven't tested this, but I suspect this is currently not working and will start working again once this PR has been merged.

## Test instructions
* (De-)Activating the plugin should still work as expected.
* The settings link for this plugin should now show up on the plugins page.

### DIFF
--- a/classes/class-wpseo-news.php
+++ b/classes/class-wpseo-news.php
@@ -191,7 +191,7 @@ class WPSEO_News {
 	public function plugin_links( $links, $file ) {
 		static $this_plugin;
 		if ( empty( $this_plugin ) ) {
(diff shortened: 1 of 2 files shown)


andizer:
CR & acceptance done :+1:

PR #362, opened 2018-02-11
CS/QA: minor regex improvements

jrfnl:
## Summary
* Fixed an issue where image information retrieved and used in the sitemap could be incomplete.

The original regex could match mis-matched quotes. The improved regexes won't.

Example:
Given the following string `alt="something he's said"`, the original regex would match `alt="something he'`, the improved regex will match `alt="something he's said"` correctly.

Given the following string `alt="something he\"s said"`, the original regex would match `alt="something he\"`, the improved regex will match `alt="something he\"s said"` correctly.

## Test instructions
* Start with the `trunk` branch
* Add a post with an image with a single/double qoute in the alt/title tag.
* Open the news image sitemap and take note of the incorrect information in the `<image:title>` and `<image:caption>` tags.
* Switch to this branch.
* Clear the sitemap cache.
* Open the new image sitemap and see the difference.


moorscode:
The improvement made on the use of the starting quote to match the ending quote is a great one.
Escaping the quote with a backslash is not supported in HMTL, thus we should not work around that functionality.

The browser will stop rendering after the`\"` in the following example: `alt="something he\"s said"`

jrfnl:
@moorscode I've added a second commit removing the bit in the regex that allowed for slash escaped quotes. Feel free to squash the commits on merge.

@jcomack Just checking - I'd previously changed the label of this PR to `needs-code-review` as it involves something which can be regarded as a _bug_ and the fix changes existing functionality in the plugin.
Also, this is not something which is checked by CS tools.
Why was this re-tagged as `yoast-cs` ?

classes/class-sitemap-images.php
```diff
@@ -78,7 +78,7 @@ private function get_item_images() {
 	 */
 	private function get_images_from_content( $matches ) {
 		foreach ( $matches[0] as $img ) {
-			if ( ! preg_match( '/src=("|\')([^"|\']+)("|\')/', $img, $match ) ) {
+			if ( ! preg_match( '`src=(["\'])(.+?)\1`', $img, $match ) ) {
```
moorscode: The -any character- should not be -one or more/`+`- as empty alt attributes will have a failing match.
Changing this to `(.*?)` allows for empty alt tags to be used.

_Empty alt tags are the default generated by WordPress when inserting a Media item in the editor_

jrfnl: Happy to make the change, just want to note that that changes the behaviour of the regex though, as previously it did demand one or more: `([^"|\']+)`.

jrfnl:
> Do you like me/us to pick that up or do you want to add them?

I'll have a go at it.

classes/class-sitemap-images.php
jrfnl: @moorscode Looking into this now, but I don't actually understand why you'd want to match empty alt tags. If the alt tag is empty, it won't be added to the sitemap anyway, so why store it ?

moorscode: That sounds logical, could you add that context `If the alt tag is empty, it won't be added to the sitemap` to the line as a comment, to make sure people will understand why and how?

jrfnl: @moorscode Will do when I get back to this PR.

jrfnl:
@moorscode Maybe we should tag this as `blocked` for now, as there are bugs in another method which prevent the unit tests I've written to pass at the moment. Those bugs will need to be addressed (in a separate PR) before this can be merged.

moorscode:
@jrfnl should this still be blocked?

jrfnl:
@jip AFAICS Yes.
As discussed above:
-> this needs unit tests,
    -> to add the unit tests for the method and have them be valid to show the difference between before/after, unit tests for the other method called within the same method need to be added too,
        -> for those to pass, some other bug fixes are needed
            -> and that's IIRC when it was decided to not spend dev time on this repo are as it was to be integrated elsewhere.

Let me know if any of that has changed ...

PR #367, opened 2018-02-13
Bugfix timezone identifier retrieval

jrfnl:
## Summary

* _N/A_ (as long as the fix is merged in the same release as the original PR which broke things, i.e. `7.1`)

This was broken by PR #359 which changed the `in_array()` to use a strict comparison as attested by the related unit test failing (see [Travis build logs](https://travis-ci.org/Yoast/wpseo-news/jobs/340512987#L696)).

In the original code, the `$timezone_option !== ''` was superfluous: it would always pass as `$timezone_option` was an `object` being strict compared to a string.

The loose `in_array()` however triggered the `__toString()` method in the `WPSEO_News_Sitemap_Timezone` class which allowed the timezone to be found in the identifiers list.

As #359 changed this to a strict comparison, that part of the condition also broke, so now the condition would always fail.

Looking at the complete function, the option retrieval is cached using the `static $timezone_option` variable, however, the output of the _identifier retrieval_ after that will never change between function calls (as the option is cached), so can be/should be cached as well.

To that end, I've refactored the method to facilitate this. The method should now be marginally faster, which for large sitemaps should make a slight difference.

## Test instructions

* This is already tested by the existing unit test which now passes again.


jcomack:
Acceptance done 👍

PR #369, opened 2018-02-14
Changed private method to public to determine timezone

### LINKED ISSUE #368: Strict type checking results in faulty datetime in Sitemap
Due to a strict type checking introduced in https://github.com/Yoast/wpseo-news/pull/359, all builds on `trunk` currently fail. This is because the expected DateTime that should be present in the sitemap items, but isn't.

jcomack:
## Summary

* Fixes an issue where an incorrect DateTime was returned due to strict checking being introduced on arrays. 

* Because the class that determines what timezone to use in the sitemap leans heavily on the `__toString()` method, it resulted in a breakage when strict checking is used on `in_array`. Because of this, I altered the method that determines the timezone and made it public so we can properly check against the expected output.

## Test instructions

* See that Travis succeeds

_-OR-_

* While using the `trunk` branch, create a new post and go to your News Sitemap.
* Notice that there's no timestamp behind the newly created post, but just the date.
* Checkout this branch.
* Create a new post.
* Go back to the News Sitemap.
* Check that there's also a timestamp present in the sitemap.

Fixes #368

### DIFF
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -251,7 +251,7 @@ class WPSEO_News_Sitemap_Item {
 		}
 
 		// Is there a timezone option and does it exists in the list of 'valid' timezone.
-		if ( $timezone_option !== '' && in_array( $timezone_option, DateTimeZone::listIdentifiers(), true ) ) {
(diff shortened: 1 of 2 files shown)


jrfnl:
Duplicate/alternative solution to of #367

jcomack:
Closing this in favor of #367.

PR #372, opened 2018-02-15
CS/XSS: always escape output [1]

jrfnl:
## Summary

_Security hardening_

* No functional changes.
* QA/CS compliance.

All output should be escaped, including translations.

Simple non-controversial ones. Use `esc_html()`, `esc_attr()` or `esc_url()` where appropriate.
These should not give any problems.

## Test instructions

Test this by checking that the meta values of a news post are still added correctly to the header of the post on the front-end.

### DIFF
--- a/classes/class-head.php
+++ b/classes/class-head.php
@@ -53,7 +53,7 @@ class WPSEO_News_Head {
 
 			$meta_news_keywords = new WPSEO_News_Meta_Keywords( $this->post->ID );
 			if ( ! empty( $meta_news_keywords ) ) {
-				echo '<meta name="news_keywords" content="' . $meta_news_keywords . '" />' . "\n";
+				echo '<meta name="news_keywords" content="' . esc_attr( $meta_news_keywords ) . '" />' . "\n";
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


classes/class-head.php
```diff
@@ -72,12 +72,12 @@ private function display_original_source() {
...
 			$original_source = trim( WPSEO_Meta::get_value( 'newssitemap-original', $this->post->ID ) );
 			if ( empty( $original_source ) ) {
-				echo '<meta name="original-source" content="' . get_permalink( $this->post->ID ) . '" />' . "\n";
+				echo '<meta name="original-source" content="' . esc_url( get_permalink( $this->post->ID ) ) . '" />' . "\n";
 			}
 			else {
 				$sources = explode( '|', $original_source );
 				foreach ( $sources as $source ) {
-					echo '<meta name="original-source" content="' . $source . '" />' . "\n";
+					echo '<meta name="original-source" content="' . esc_attr( $source ) . '" />' . "\n";
```
jdevalk: you're using `esc_url` above, since this should always be a URL as well, should we use `esc_url` here, or a combination of `esc_attr` and `esc_url`?

jrfnl: I didn't realize that `$source` would always be a URL. In that case, using `esc_url()` should be sufficient. Changing this now.

jrfnl: Done. As the change was so small, I've just amended the commit.

PR #373, opened 2018-02-15
CS/XSS: always escape output [2]

jrfnl:
## Summary

_Security hardening_

* No functional changes.
* QA/CS compliance.

All output should be escaped, including translations.

Includes minor code build up/layout changes.

## Test instructions

Test this by checking that the admin notices which display when WPSEO is missing or not at the correct version and when WP is not at the correct minimum version still display correctly.


jrfnl:
Rebased for merge conflicts.

PR #376, opened 2018-02-15
CS/XSS: remove need to escape output [5]

jrfnl:
## Summary

_Security hardening_

* No functional changes.
* QA/CS compliance.

While I understand that having this string in a variable is useful, there isn't any real need. It's not as if the variable is used in a lot of places throughout the code. Just few times directly following the old variable definition, so removing the variable was simplest.

Alternatively this block could be whitelisted, but that would open it up to new issues being introduced at a later stage.

## Test instructions

Test this by checking that the `editors pick` section of the admin page still displays correctly as expected.


andizer:
CR & acceptance done

PR #385, opened 2018-03-14
CS/QA: don't unnecessarily use `readfile()` function call

jrfnl:
## Summary
_N/A_

* No functional changes.

No need to use a file system function when a simple language construct should have the same effect and will be more efficient.

## Test instructions
* Make sure that the xsl file is not reachable directly by the server - i.e. this can be done via an `.htaccess` entry to direct requests to `.xsl` files to WP.
* Clear out the sitemap cache
* Load the sitemap and verify that the XSL stylesheet is correctly served with it.


andizer:
CR & Acceptance done

Issue #402, opened 2018-05-14
Travis script needs review as to the running of the unit tests


jrfnl:
There is something weird going on with the Travis script as the build for PR #401 should fail on **all** PHP versions when run against the second commit, but only the PHP7 versions fail: https://travis-ci.org/Yoast/wpseo-news/builds/378697844

I've tested locally against various PHP 5 versions and the unit tests fail correctly in that case.

As due to the folding, I can't actually see whether the tests are run at all, this is now really difficult to debug.

What I suspect is that the unit tests are currently **not** being run against PHP 5.

/cc @moorscode

moorscode:
I see they are being run, and that a `Fatal` error is occuring, but the status-code of the run is not being caught as expected.

jrfnl:
Hmm... will you handle this one ? I.e. I can let go of it ?

moorscode:
I'll put this in the process for somebody to pick this up.

PR #403, opened 2018-05-14
QA: skip test when minimum PHP version does not comply

jrfnl:
## Summary

* _N/A_

The `DateTimeImmutable` class was only introduced in PHP 5.5.0.
The `DateInterval` class in PHP 5.3.0.

This test can therefore not be run on PHP < 5.5.0.

Adding the PHPUnit `@requires` annotation will ensure that the test suite will not break on this test.

Alternatively, if testing this functionality on PHP 5.2 - 5.4 is desired, the test will need to be refactored and the above mentioned classes removed from the test.

Ref:
* https://phpunit.de/manual/4.8/en/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests.requires.tables.api

## Test instructions

* Locally switch to PHP 5.4 and the `trunk` branch, run the unit tests and see them fail with a fatal PHP error.
* Switch to this branch and see the unit test skip the test in question.

The Travis build should have alerted about this when the original unit test were pulled, but didn't due to  issue #402.

### DIFF
--- a/tests/test-class-sitemap.php
+++ b/tests/test-class-sitemap.php
@@ -257,6 +257,8 @@ class WPSEO_News_Sitemap_Test extends WPSEO_News_UnitTestCase {
 	 * Checks that expired posts don't get included in the sitemap.
 	 *
 	 * @covers WPSEO_News_Sitemap::build_sitemap
+	 *
+	 * @requires PHP 5.5
 	 */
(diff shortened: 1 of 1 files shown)


jcomack:
Closing this in favor of https://github.com/Yoast/wpseo-news/pull/405 as it doesn't seem wise to be to skip tests.

PR #416, opened 2018-07-04
Remove original source

jdevalk:
## Summary

* Removes functionality for original-source as Google has confirmed they're no longer using it.

* Deleted all code that had to do with it, didn't deprecate because there's no need. There were no tests for this.
* Deleted the data that was no longer useful, also deleted the data for functionality that was removed in 7.7.
* Decided to _keep_ the meta news keywords fields, people might be using those for something else and it was editorially put in, so I don't really want to delete it without user consent.

## Test instructions

* Go to News SEO settings on a post, see original source is gone.

### DIFF
--- a/classes/class-upgrade-manager.php
+++ b/classes/class-upgrade-manager.php
@@ -24,13 +24,11 @@ class WPSEO_News_Upgrade_Manager {
 
 		// Check if update is required.
 		if ( version_compare( WPSEO_News::VERSION, $options['version'], '>' ) ) {
-
 			// Do update.
 			$this->do_update( $options['version'] );
 
 			// Update version code.
 			$this->update_current_version_code();
-
 		}
 
 	}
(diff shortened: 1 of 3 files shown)


classes/class-upgrade-manager.php
```diff
@@ -84,4 +63,51 @@ private function update_current_version_code() {
...
+		$news_options = null;
+	}
+
+	/**
+	 * Perform the upgrade to 7.8.
+	 */
+	private function upgrade_78() {
+		global $wpdb;
+		// Delete all standout tags. Functionality was deleted in 7.7, data only deleted in 7.8.
+		$wpdb->query( "DELETE FROM $wpdb->postmeta WHERE meta_key = '_yoast_wpseo_newssitemap-standout'" );
```
andizer: I would choose for:

```
$wpdb->delete( 
    $wpdb->postmeta, 
    array(
       'meta_key' => '_yoast_wpseo_newssitemap-standout',
    ), 
    array( '%s' )
);
```

jdevalk:
Changes made, ready for another review :)

PR #427, opened 2018-09-14
Make sure the artifact matches the release zip

### LINKED ISSUE #426: Fix artifact configuration
The current configuration of the artifact creation via `grunt` has errors.

The artifact file does not contain a folder and from the last experience on another repository I'm quite sure that there will be files missing or included that should not be there.

The zip created using `grunt artifact` needs to match very closely with the one created with the `create_beta.sh` script in the Wiki repository.

Reference:
https://github.com/Yoast/wordpress-seo/blob/trunk/grunt/config/compress.js#L11

Only after this has been applied and verified can the configuration of News SEO in the https://github.com/Yoast/Wiki/blob/master/includes/product_choice.sh#L76 be changed to include `GRUNT_ARTIFACT="1"`

moorscode:
This fixes the grunt tasks to build a zip file that matches the release zip
Removes abundant grunt commands
Uses the general "artifact" for consistency and Wiki build script compatibility

**After this is merged, the release and beta script need to be updated** - when the release branch is made that includes this change! (Suspected to be 8.4)
For instructions, see issue.

## Summary

* _N/A_

*

## Test instructions

* Download the latest zip of News SEO (from the releases on this repo)
* Run `grunt artifact` in the branch
* Verify that all files and folders you expect to be there are actually in the `artifact.zip`
* Verify that the zip size closely matches the release zip (it is slightly bigger because certain files in the `vendor` folder are not specifically removed)

Fixes #426

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -22,6 +22,7 @@ module.exports = function( grunt ) {
 			logs: "logs/"
 		},
 		files: {
+			artifact: "artifact",
 			js: [
 				"assets/**/*.js",
 				"!assets/**/*.min.js"
--- a/grunt/config/aliases.yaml
+++ b/grunt/config/aliases.yaml
@@ -4,6 +4,7 @@
 build:
   - 'build:js'
   - 'build:i18n'
+
 # Build JavaScript from assets to production
 'build:js':
   - 'uglify'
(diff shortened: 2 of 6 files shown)


moorscode:
Outdated.

Issue #438, opened 2018-10-05
Sitemap contains image attachment GUIDs instead of file URLs.


ChristophKnopf:
### What did you expect to happen?
Correct image URLs <image:loc> of posts in our news-sitemap.xml e.g.:
```
<url>
	...
	<image:image>
		<image:loc>https://bnn.de/wp-content/uploads/2017/08/blaulicht-rss.jpg</image:loc>
		<image:title>blaulicht-rss</image:title>
	</image:image>
</url>
```

### What happened instead?
Google News sitemap contains image attachment GUIDs instead of file URLs, which results in broken image URLs in news-sitemap.xml:
```
<image:loc>http://dpa/weblines/../publishedPictures/large/urn-newsml-dpa-com-20090101-181005-99-248701_large_4_3.jpg</image:loc>
```

### How can we reproduce this behavior?
Import images with uuids of an external service and you will have it.

Can you provide a link to a page which shows this issue?
view-source:https://bnn.de/news-sitemap.xml

### Technical info
* WordPress version: 4.9.7
* Yoast SEO version: 8.3

Proposed solution:
Change https://github.com/Yoast/wpseo-news/blob/trunk/classes/class-sitemap-images.php#L217-L222 to
```
if ( ! empty( $attachment['href'] ) ) {
  $this->images[ $attachment['href'] ] = $image;
}
elseif ( ! empty( $attachment['src'] ) ) {
  $this->images[ $attachment['src'] ] = $image;
}
```
and https://github.com/Yoast/wpseo-news/blob/trunk/classes/class-sitemap-images.php#L245 to
`'href'        => wp_get_attachment_url( $attachment->ID ),`

jono-alderson:
Closing as fixed

PR #442, opened 2018-10-26
Fixed Sitemap image URLs to use href instead of guid.

ChristophKnopf:
## Summary

* Fixed Sitemap image URLs to use href instead of guid to prevent relative URLs in the Sitemap.

* Relative image URLs in the Sitemap which are not interpreted correctly by google. Besides not SEO friendly.

## Test instructions

* https://github.com/Yoast/wpseo-news/issues/438

Fixes #

### DIFF
--- a/classes/class-sitemap-images.php
+++ b/classes/class-sitemap-images.php
@@ -214,11 +214,11 @@ class WPSEO_News_Sitemap_Images {
 			$image['alt'] = $attachment['alt'];
 		}
 
-		if ( ! empty( $attachment['src'] ) ) {
-			$this->images[ $attachment['src'] ] = $image;
-		}
-		elseif ( ! empty( $attachment['href'] ) ) {
+		if ( ! empty( $attachment['href'] ) ) {
 			$this->images[ $attachment['href'] ] = $image;
+		}
+		elseif ( ! empty( $attachment['src'] ) ) {
+			$this->images[ $attachment['src'] ] = $image;
 		}
 	}
 
@@ -242,7 +242,7 @@ class WPSEO_News_Sitemap_Images {
 		return array(
 			'title'       => $attachment->post_title,
 			'alt'         => get_post_meta( $attachment->ID, '_wp_attachment_image_alt', true ),
-			'href'        => get_permalink( $attachment->ID ),
+			'href'        => wp_get_attachment_url( $attachment->ID ),
 			'src'         => $attachment->guid,
 		);
 	}


Xyfi:
Hi @ChristophKnopf, first and foremost thank you for your PR. 

The test instructions aren't clear. I'm not sure how I can reproduce the issue in order to confirm your PR. Could you please confirm the issue is still relevant and update your test instructions with more concise steps?

ChristophKnopf:
Hi @Xyfi,

sorry for the late response. It's been a while but if I remember correctly the problem occurs when we import  3rd party data (XML) which contains relative URLs for images.

The URLs are processed this way until the end, so in the stiemap.xml we will have something like this:
```
<url>
  <loc>https://bnn.de/nachrichten/wirtschaft/eine-million-e-autos-regierungsberater-schieben-ziel</loc>
  <lastmod>2018-09-19T16:25:43+02:00</lastmod>
  <image:image>
    <image:loc>../uploads/2018/09/dpa-urn-newsml-dpa-com-20090101-180919-99-20662_large_4_3.jpg</image:loc>
    <image:title><![CDATA[urn-newsml-dpa-com-20090101-180919-99-20662_large_4_3.jpg]]></image:title>
    <image:caption><![CDATA[Parkplatz für Elektroautos]]></image:caption>
  </image:image>
</url>
```
the tag `<image:loc>` contains a relative URL because the third party data contains relative URLs and this is processed through `get_permalink( $attachment->ID )` until the end. If you use use `wp_get_attachment_url( $attachment->ID )` instead, it doesn't matter anymore if the URLs are relative or not, especially when you test for `$attachment['href']` first, right?

So test your output with a relative URL and you will the the same result in the sitemap. Did that help?

ChristophKnopf:
Great 😞

ChristophKnopf:
Sorry fired the wrong button 😢

sun:
More precisely, the problem lays here:
https://github.com/Yoast/wpseo-news/blob/d1ee38faba22cc3e27d53cd77a0e2f6ee11a6635/classes/class-sitemap-images.php#L246

The value in `wp_posts.guid` is a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier), which _can_ be a URL, or something that looks like a URL, but does not have to.

In content import scenarios (very typical for content-driven news sites), the UUID is used to record the original identifier of a document, so as to find the already imported content again in case an update for it is to be imported.  For example:
```
http:://dpa/infoline/urn-newsml-dpa-com-20090101-190117-99-600994.jpg
```
Whereas "dpa" refers to the origin.

WordPress Core happens to store a value that looks like the URL in `guid`, but it is not.  It represents a UUID v5, which can be a regular URI.  The value is not maintained nor used any further for output purposes by WordPress Core itself.  Instead, WordPress always uses the proper API functions for generating a URL for an image attachment (like this PR does).

The only valid use for a UUID in a feed would be the feed item's identifier. But as explained in the above, it cannot be used for outputting a post's image URL.

Therefore, the usage of `src` in `get_item_featured_image()` should actually be removed altogether:

https://github.com/Yoast/wpseo-news/blob/d1ee38faba22cc3e27d53cd77a0e2f6ee11a6635/classes/class-sitemap-images.php#L217-L222

Doing so leaves us with `href`, which this PR corrects to generate the proper URL of the image attachment like WordPress Core does in all other situations to create image URLs.

If you agree, @ChristophKnopf can remove the lines using `src` instead of switching their order in usage.

andizer:
Can you tell us what you mean by importing third party data? Do you have an example?

ChristophKnopf:
Sure, here you go (shortened source), have a look at the `media-reference`.
```
<?xml version="1.0" encoding="UTF-8"?>
<!-- DOCTYPE nitf PUBLIC "-//IPTC-NAA//DTD NITF-XML 3.0//EN" "nitf.dtd" -->
<nitf> 
  <head> 
    <title>Netztechnik nur noch von «vertrauenswürdigen Lieferanten»</title>  
    <meta content="dpa - Deutsche Presse-Agentur GmbH" name="origin"/>  
    <meta content="dpa-infocom GmbH" name="copyright"/>  
    <meta content="service-wds" name="generator"/>  
    <docdata> 
      <evloc iso-cc="DEU"/>  
      <doc-id id-string="urn-newsml-dpa-com-20090101-190307" regsrc="dpa-infocom"/>  
      <urgency ed-urg="4"/>  
      <fixture fix-id="/infoline/netzwelt/"/>  
      <date.issue norm="20190307T190437+0100"/>  
      <doc-scope scope="wi"/>  
      <key-list> 
        <keyword key="Telekommunikation/Internet/5G/Regulierung/Huawei/Spionageverdacht/Deutschland/China/"/> 
      </key-list> 
    </docdata> 
  </head>  
  <body> 
    <body.head> 
      <hedline> 
        <hl1>Netztechnik nur noch von «vertrauenswürdigen Lieferanten»</hl1>  
        <hl2 class="ShortHeadLine">Kommt eine «Lex Huawei»?</hl2> 
      </hedline>  
      <abstract> 
        <p>Am Fall Huawei ...</p> 
      </abstract> 
    </body.head>  
    <body.content>
      <p>Bonn (dpa) - In Deutschland zeichnen ...</p>
      <media media-type="image"> 
        <media-metadata name="media-id" value="urn-newsml-dpa-com-20090101-190307-99-285264:1551981717000"/>  
        <media-metadata name="image-fovea-region" value="41,151,468,451"/>  
        <media-reference alternate-text="Huawei" height="113" mime-type="image/jpeg" name="small_16_9" source="../publishedPictures/small/urn-newsml-dpa-com-20090101-190307-99-285264_small_16_9.jpg" width="200"/>  
        <media-reference alternate-text="Huawei" height="150" mime-type="image/jpeg" name="small_4_3" source="../publishedPictures/small/urn-newsml-dpa-com-20090101-190307-99-285264_small_4_3.jpg" width="200"/>  
        <media-reference alternate-text="Huawei" height="600" mime-type="image/jpeg" name="large_4_3" source="../publishedPictures/large/urn-newsml-dpa-com-20090101-190307-99-285264_large_4_3.jpg" width="800"/>  
      </media> 
    </body.content>  
    <body.end/> 
  </body> 
</nitf>
```

sun:
To explain this further, in the example data shared by @ChristophKnopf we see the following identifiers in the XML elements:
```xml
<meta content="dpa - Deutsche Presse-Agentur GmbH" name="origin" />
...
<fixture fix-id="/infoline/netzwelt/" />
...
<media-metadata name="media-id"
 value="urn-newsml-dpa-com-20090101-190307-99-285264:1551981717000" />
```
In order to uniquely identify the media file of this content now and forever in the future, the import generates the following UUID:
```
http://<origin>/<fixture>/<media-id>

http://dpa/infoline/urn-newsml-dpa-com-20090101-190307-99-285264.jpg
```
The UUID needs to be manually generated in this case, because the content provider does not provide an actual UUID for each media file.  Other content providers do provide their own UUID, which must be taken over into all remote systems that are storing copies of a document or file, so as to be able to uniquely identify it in case an update for it is pushed.

In any case, the UUID is not an actual URL that can be used for referencing the source URL of an image on the web.  To generate a public URL for an image, the proper WordPress API function needs to be used.

IreneStr:
Acceptance 👍

PR #446, opened 2018-11-01
Use the SEO title instead of default WordPress title for News Items

timnolte:
## Summary

* Changes news items to use the SEO title instead of default WordPress title.

* Added a method that performs the same sort of treatment of using the default plugin options that is used when building the new items.

## Test instructions

* news-sitemap.xml should display either a post-types custom SEO title or the default plugin title format.

Fixes #64

### DIFF
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -192,18 +192,23 @@ class WPSEO_News_Sitemap_Item {
 	 *
 	 * @param   WP_Post $item The post object.
 	 *
-	 * @return  string        The WPSEO formatted title or, if problems, the post_title.
+	 * @return  string	The formatted title or, if no formatted title can be created, the post_title.
 	 */
 	protected function get_item_title( $item = null ) {
-		// Custom WPSEO post type title.
+		// Exit early if the item is null.
+		if ( $item === null ) {
+			return '';
+		}
+
+    // Custom WPSEO post type title.
 		$title = WPSEO_Meta::get_value( 'title', $item->ID );
-		if ( $title != '' && false !== $title ) {
+		if ( $title !=== '' && $title !=== false ) {
 			return wpseo_replace_vars( $title, $item );
 		}
 
 		// Default WPSEO post type title.
 		$defaults = WPSEO_Option_Titles::get_instance()->get_defaults();
-		if ( array_key_exists( 'title-' . $item->post_type, $defaults ) && false !== $defaults ) {
+		if ( array_key_exists( 'title-' . $item->post_type, $defaults ) && $defaults !=== false ) {
 			return wpseo_replace_vars( str_replace( ' %%page%% ', ' ', $defaults[ 'title-' . $item->post_type ] ), $item );
 		}
 
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -200,7 +200,7 @@ class WPSEO_News_Sitemap_Item {
 			return '';
 		}
 
-    // Custom WPSEO post type title.
+		// Custom WPSEO post type title.
 		$title = WPSEO_Meta::get_value( 'title', $item->ID );
 		if ( $title !== '' && $title !== false ) {
 			return wpseo_replace_vars( $title, $item );
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -151,9 +151,9 @@ class WPSEO_News_Sitemap_Item {
 	 */
 	private function build_news_tag() {
 
-    $title         = $this->get_item_title();
-    $genre         = $this->get_item_genre();
-    $stock_tickers = $this->get_item_stock_tickers( $this->item->ID );
+		$title         = $this->get_item_title();
+		$genre         = $this->get_item_genre();
+		$stock_tickers = $this->get_item_stock_tickers( $this->item->ID );
 
 		$this->output .= "\t<news:news>\n";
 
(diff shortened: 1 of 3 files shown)


classes/class-sitemap-item.php
```diff
@@ -186,7 +187,27 @@ private function build_publication_tag() {
 		$this->output .= "\t\t</news:publication>\n";
 	}
 
-	/**
+  /**
+   * Gets the SEO title of the item.
+   *
+   * @return string
```
jcomack: Missing `@return` documentation.

classes/class-sitemap-item.php
```diff
@@ -186,7 +187,27 @@ private function build_publication_tag() {
...
+  private function get_item_title() {
+    $title = WPSEO_Meta::get_value( 'title', $this->item->ID, true );
+    
+    if ( empty( $title ) ) {
+      $default_from_options = WPSEO_Options::get_default( 'wpseo_titles', $this->item->post_type );
+      if ( false !== $default_from_options ) { 
+        $title = str_replace( ' %%page%% ', ' ', $default_from_options );
+      }
+    } else {
+      $title = '%%title%%';
```
jcomack: Something seems terribly strange about this line. If I understand correctly, if the title is set within `WPSEO_Meta`, this line overrides the `$title` variable again. Not sure why you'd want that?

classes/class-sitemap-item.php
```diff
@@ -186,7 +187,27 @@ private function build_publication_tag() {
 		$this->output .= "\t\t</news:publication>\n";
 	}
 
-	/**
+  /**
+   * Gets the SEO title of the item.
+   *
+   * @return string
+   */
+  private function get_item_title() {
```
jcomack: I'd suggest making this a protected function that passes the item as a parameter (i.e. `$item`) to make this function testable.

classes/class-sitemap-item.php
```diff
@@ -186,7 +187,27 @@ private function build_publication_tag() {
...
+    $title = WPSEO_Meta::get_value( 'title', $this->item->ID, true );
+    
+    if ( empty( $title ) ) {
+      $default_from_options = WPSEO_Options::get_default( 'wpseo_titles', $this->item->post_type );
+      if ( false !== $default_from_options ) {
```
jcomack: We tend to not use the Yoda-style if-statements.

classes/class-sitemap-item.php
```diff
@@ -186,7 +187,27 @@ private function build_publication_tag() {
...
-	/**
+  /**
+   * Gets the SEO title of the item.
+   *
+   * @return string
+   */
+  private function get_item_title() {
+    $title = WPSEO_Meta::get_value( 'title', $this->item->ID, true );
+    
+    if ( empty( $title ) ) {
```
jcomack: Turning this into an early return by inverting the if will declutter the nested if-statements, making the code more readable.

classes/class-sitemap-item.php
timnolte: hmm, OK, I can address that, my preference is to stay away from `=` checks due to the rash of typos causing assignments.

classes/class-sitemap-item.php
timnolte: Good call, I was too quick to follow some older WPSEO code. Addressing.

classes/class-sitemap-item.php
timnolte: Oye, yes, that is a total missing on my part. Fixing.

classes/class-sitemap-item.php
timnolte: Will add, thanks.

timnolte:
I have a method call to fix on this I see. Addressing.

timnolte:
This PR is not ready. The SEO Title is being used for custom SEO titles but not for any that use the default plugin titles. WPSEO_Meta::get_value() is not returning anything but the custom SEO titles, which is find but the WPSEO_Options::get_default is not returning the expected default SEO title format. The PR is passing unit tests because the fallback is to use the WordPress post_title as before when the first 2 tries fail.

timnolte:
The code is working as intended and formatting should be good. Looks like I need to update some unit tests to account for using the WPSEO titles. Do we need a new plugin setting to turn on/off this feature?

timnolte:
@jcomack I believe everything is in order now. I do have the outstanding question of whether I need to add a plugin setting for users to be able to turn on/off this feature? I will need to update unit test, and add, if a plugin setting is needed. Thanks!

jcomack:
@timnolte It'd make more sense to use the proper SEO title instead of the WordPress title, so adding a feature flag for this seems unnecessary to me. I'm going to subject this code to a second code-review + ask if my assumptions regarding the toggling of this feature is correct and report back :)

classes/class-sitemap-item.php
```diff
@@ -186,6 +187,30 @@ private function build_publication_tag() {
...
+	 * Gets the SEO title of the item.
+	 *
+	 * @param   WP_Post $item The post object.
+	 *
+	 * @return  string        The WPSEO formatted title or, if problems, the post_title.
+	 */
+	protected function get_item_title( $item = null ) {
+		// Custom WPSEO post type title.
+		$title = WPSEO_Meta::get_value( 'title', $item->ID );
+		if ( $title != '' && false !== $title ) {
```
jcomack: Please apply strict checking for `$title` (i.e. `$title !=== ''`).

jcomack: Also please keep the checking of the variable consistent (i.e. `$title === false` instead of `false === $title`).

classes/class-sitemap-item.php
```diff
@@ -186,6 +187,30 @@ private function build_publication_tag() {
...
+			return wpseo_replace_vars( $title, $item );
+		}
+
+		// Default WPSEO post type title.
+		$defaults = WPSEO_Option_Titles::get_instance()->get_defaults();
+		if ( array_key_exists( 'title-' . $item->post_type, $defaults ) && false !== $defaults ) {
```
jcomack: Please inverse `false !== $defaults` to `$defaults !== false`.

jcomack:
I've checked and we don't find a toggle necessary.

classes/class-sitemap.php
```diff
@@ -222,7 +222,7 @@ private function get_items( $limit = 1000 ) {
 				"SELECT ID, post_content, post_name, post_author, post_parent, post_date_gmt, post_date, post_date_gmt, post_title, post_type
 				FROM {$wpdb->posts}
 				WHERE post_status='publish'
-					AND ( TIMESTAMPDIFF( MINUTE, post_date_gmt, NOW() ) <= ( 48 * 60 ) )
+					AND ( TIMESTAMPDIFF( MINUTE, post_date_gmt, UTC_TIMESTAMP() ) <= ( 48 * 60 ) )
```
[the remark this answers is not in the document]
timnolte: NOW() is using the database configured, or server, timezone and not GMT, need to ensure the correct posts are being returned.

classes/class-sitemap-item.php
```diff
@@ -186,6 +187,30 @@ private function build_publication_tag() {
...
+	/**
+	 * Gets the SEO title of the item.
+	 *
+	 * @param   WP_Post $item The post object.
+	 *
+	 * @return  string        The WPSEO formatted title or, if problems, the post_title.
+	 */
+	protected function get_item_title( $item = null ) {
+		// Custom WPSEO post type title.
+		$title = WPSEO_Meta::get_value( 'title', $item->ID );
```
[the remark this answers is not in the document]
timnolte: ah, good catch, adding handling for that.

classes/class-sitemap-item.php
timnolte: So, I 2nd guessed myself when I saw you have !=== and made the mistake of taking that literally, no it's !==.

classes/class-sitemap-item.php
```diff
@@ -186,6 +187,35 @@ private function build_publication_tag() {
...
+	 * @return  string	The formatted title or, if no formatted title can be created, the post_title.
+	 */
+	protected function get_item_title( $item = null ) {
+		// Exit early if the item is null.
+		if ( $item === null ) {
+			return '';
+		}
+
+		// Custom WPSEO post type title.
+		$title = WPSEO_Meta::get_value( 'title', $item->ID );
```
andizer: I would suggest to use `WPSEO_Frontend::get_instance()->get_content_title( $item )` which already does the same as the content of the method.

IreneStr:
Replaced by #455

PR #450, opened 2018-11-27
Use UTC for the sitemap index

### LINKED ISSUE #435: Wrong date for news sitemap in sitemap index
### What did you expect to happen?
The sitemap index would list the correct date format.

### What happened instead?
Wrong date format

Install Yoast News 8.3
Set WordPress time zone to UTC
Publish a post
Compare the dates from the index and the newest post in the news sitemap. Correct UTC time on both index and news sitemap.
Change WP time zone to something else.
Disable/Enable the sitemap feature.
Compare the dates from the index and the newest post in the news sitemap. Same date/time but the index still includes the time zone designator which makes it the wrong time.

### Screenshots
News sitemap uses UTC/GMT
selection_200

Index uses UTC/GMT time but wrong appends time zone designator
selection_199

### Technical info
* WordPress version: 4.9.8
* Yoast SEO version: 8.3

Related to recent change? https://github.com/Yoast/wpseo-news/issues/227

andizer:
## Summary

* Fixes a bug where the last modified date for the sitemap index has been displayed with the wrong time zone 

* I've chosen to use UTC, because we are relying on the GMT date for the last modified date.

## Test instructions

* Follow the steps from issue #435:
> Install Yoast News 8.3
> Set WordPress time zone to UTC
> Publish a post
> Compare the dates from the index and the newest post in the news sitemap. Correct UTC time on both index and news sitemap.
> Change WP time zone to something else.
> Disable/Enable the sitemap feature.
> Compare the dates from the index and the newest post in the news sitemap. Same date/time but the index still includes the time zone designator which makes it the wrong time.

* It should be solved now

Fixes #435

### DIFF
--- a/classes/class-sitemap.php
+++ b/classes/class-sitemap.php
@@ -52,7 +52,7 @@ class WPSEO_News_Sitemap {
 			return $str;
 		}
 
-		$date = new DateTime( get_lastpostdate( 'gmt' ), new DateTimeZone( new WPSEO_News_Sitemap_Timezone() ) );
+		$date = new DateTime( get_lastpostdate( 'gmt' ), new DateTimeZone( 'UTC' ) );
 
 		$str .= '<sitemap>' . "\n";
 		$str .= '<loc>' . self::get_sitemap_name() . '</loc>' . "\n";


afercia:
CR 👍 

However, I'm not sure I fully understand if this actually solves the issue. I've published a test post at 11:17 CET (UTC + 1) and the posts sitemap index last modified date show it correctly; instead the News sitemap index shows the UTC time without offset:

screenshot 2018-11-28 at 11 28 22

I guess they should have the same time? 

Yes, the news item time now matches the index time but to my understanding it's the item time the one that is wrong: I haven't published my post at `10:17:44+00:00`

screenshot 2018-11-28 at 11 33 26

Xyfi:
I can indeed reproduce @afercia 's issue, but it does fix the original issue.

Since `2018-11-28 11:17 +01:00` and `20-11-28 10:17 +00:00` represent the same time I'm giving this a 👍  for acceptance.

PR #460, opened 2019-02-01
Fixes issue where the default SEO title was not shown in news-sitemap. 

### LINKED ISSUE #459: SEO title is not shown in news sitemap when SEO title has default replacevars
1. Activate News 9.7 beta1 plugin.
2. Make sure Search appearance for Posts is configured:
screenshot 2019-01-30 at 15 09 08

3. Create new post with title "test post".
4. Don't change anything in the snippet preview. That way, the default from search appearance will be used.
4. Check news sitemap:
screenshot 2019-01-30 at 15 12 14

It is expected to see SEO title "test post - Local WordPress Dev" in news sitemap.

6. When you add a replacevar to the title in the snippet preview, save the post, and then open the news sitemap again, the SEO title is shown in the sitemap.
schermafbeelding 2019-01-31 om 08 18 00
In this example: %%title%% %%sep%% %%sitename%% %%title%%

Issue comes from testing #455 

### Technical info
* WordPress version: 5.0.3
* Yoast SEO version: 9.7 beta1

Dieterrr:
## Summary

* N/A

* Users might see the titles in their sitemap change because of this PR because it used to only show the post_title. Now it will show the SEO title, which when left untouched outputs the default settings for SEO title set under "Search Appearance", which in its turn defaults to post_title + site name when left untouched. 
* Added an extra test to cover the scenario where a default SEO title is set and where it is not. 

## Test instructions

* Reproduce the steps in issue #459 
* Check out this branch.
* See it is fixed :)
* You can set the SEO title for a post in the "SEO title" field in the metabox (under 'edit snippet').
* The hierarchy should now function as follows: 
1) If a specific SEO title for a post is set this should be the title in the sitemap. 
2) If the SEO title for a post is set to empty the title in the news sitemap should default to the default SEO title set under `search appearance` for this post type. 
3) If this is also set to empty (so both the post seo title as well as the default seo title are empty) the regular post title is shown. 

Fixes #459

### DIFF
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -214,7 +214,7 @@ class WPSEO_News_Sitemap_Item {
 	 * Gets the SEO title of an item, used for static home and posts pages as well as singular titles.
 	 *
 	 * @codeCoverageIgnore
-	 *                    
+	 *
 	 * @param WP_Post $item Item to get the title for.
 	 *
 	 * @return string The formatted title.
--- a/classes/class-sitemap-item.php
+++ b/classes/class-sitemap-item.php
@@ -199,9 +199,8 @@ class WPSEO_News_Sitemap_Item {
 		if ( $item === null ) {
 			return '';
 		}
-
 		// Get the SEO title.
-		$title = WPSEO_Frontend::get_instance()->get_seo_title( $item );
+		$title = WPSEO_Frontend::get_instance()->get_content_title( $item );
 
 		if ( ! empty( $title ) ) {
 			return $title;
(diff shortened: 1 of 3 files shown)


classes/class-sitemap-item.php
```diff
@@ -199,9 +199,8 @@ protected function get_item_title( $item = null ) {
 		if ( $item === null ) {
 			return '';
 		}
-
 		// Get the SEO title.
-		$title = WPSEO_Frontend::get_instance()->get_seo_title( $item );
+		$title = WPSEO_Frontend::get_instance()->get_content_title( $item );
```
andizer: Have you considered wrapping this in a protected method for mocking this in testing?

PR #466, opened 2019-03-07
PHPUnit: Rename the config file

jrfnl:
## Summary
* _N/A_

> If `phpunit.xml` or `phpunit.xml.dist` (in that order) exist in the current working directory and `--configuration` is not used, the configuration will be automatically read from that file.

From: https://phpunit.de/manual/6.5/en/organizing-tests.html#organizing-tests.xml-configuration

In other words, using a `phpunit.xml.dist` file allows individual developers to overrule it with their own version if so desired.

It also means that the configuration file does not need to be passed to PHPUnit in the Travis scripts.

Includes adding `phpunit.xml` to the `.gitignore` file to allow devs to use that file for their own custom configuration.

## Test instructions

* _N/A_
    Check that the Travis builds still pass & correctly run the unit tests.


johannadevos:
CR 👍 
Acceptance 👍

PR #470, opened 2019-03-08
Prepared for PHP 5.2-5.5 drop

### LINKED ISSUE #465: Prepare for PHP 5.2-5.5 drop
Mimic [the changes that were done on wordpress-seo](https://github.com/Yoast/wordpress-seo/pull/12129). 

This includes at least:
- [x] Change `dist` from `trusty` to `xenial`. But explicitly use 'trusty' in the 5.6 test.
- [x] Add mysql as service, because it is not included by default in xenial.
- [ ] Remove multisite from the 7.3 build that doesn't run the coverage.
- [x] Remove PHP 5.3 build
- [x] Remove PHP 7.0 build
- [x] Use `WP_VERSION=latest` instead of `WP_VERSION=4.9`/`WP_VERSION=5.0`
- [x] Set the wordpress clone depth to 1 instead of 50 to speed up the build 
     - (see https://github.com/Yoast/wordpress-seo/pull/12129/files#diff-354f30a63fb0907d4ad57269548329e3L154)
- [x] Create 2 allowed failures instead of 1: one with `php: nightly` and `env: WP_VERSION=latest`, and one with `php: 7.3` and `env: WP_VERSION=master`. That way, we can distinguish between failures caused by the php build, and failures caused by the WP build. 
     - (See https://github.com/Yoast/wordpress-seo/pull/12129/files#diff-354f30a63fb0907d4ad57269548329e3R100)
- [x] Make the use of phpunit more explicit: use `if [[ "$PHPUNIT" == "1" ]]; then` instead of `if [[ "$COVERAGE" != "1" ]]; then`. 
     - See https://github.com/Yoast/wordpress-seo/pull/12129/files#diff-354f30a63fb0907d4ad57269548329e3L233. 
     - [x] Set 'PHPUNIT=1' to all builds that currently don't have 'COVERAGE=1'

hwinne:
## Summary

[More info from same changes on wordpress-seo
](https://github.com/Yoast/wordpress-seo/pull/12129)

In .travis.yml:

* Changed dist from trusty to xenial. But explicitly used 'trusty' in the 5.6 test.
* Added mysql as a service.
* Removed PHP 5.3 and 7.0 from the builds
* Removed multisite from the 7.3 build that doesn't run coverage.
* Using WP_VERSION=latest instead of WP_VERSION=4.9 / WP_VERSION=5.0
* Upgraded PHP 5.2 test from 4.8 to WP_VERSION=4.9
* WordPress Core and wordpress-seo clone depth set from 50 to 1
* Created 2 allowed failures
* Added a script which gets the latest WordPress version

* As WPSEO News is not upgraded to PHP 7.3 yet, I could not remove multisite (3rd checkbox in isue).

* I added a scripts which gets the latest WordPress version, `in the before scripts:`:
```
# Careful: The HTTPS version of the following URL is different, therefore we need to use HTTP.
 - |
   if [[ "$WP_VERSION" == "latest" ]]; then
     curl -s http://api.wordpress.org/core/version-check/1.7/ > /tmp/wp-latest.json
     WP_VERSION=$(grep -o '"version":"[^"]*' /tmp/wp-latest.json | sed 's/"version":"//')
   fi 
```

## Test instructions

* Travis build doesn't fail
* Travis build doesn't test on PHP5.2.-5.5 anymore.
* No plugin related functionalities should be changed or behave differently.

Fixes #465


IreneStr:
I'm merging this as a first iteration. More changes are needed.

PR #477, opened 2019-03-11
CS: Rename non-snakecase variable

jrfnl:
## Summary
* _N/A_

* No functional changes.
* Code style compliance.

Variable names should be in `snake_case`.

Note: Local variable names can be safely changed to comply with this rule without breaking anything.

## Test instructions

* _N/A_
    This is a code-only change and should have no effect on the functionality.


johannadevos:
👍

PR #490, opened 2019-04-03
CS/QA: make sure $_SERVER array key is set before using it

jrfnl:
## Summary
* _N/A_

Depending on server configuration, the `REQUEST_URI` for the `$_SERVER` superglobal _may_ not be available. This prevents an "Undefined index" notice if that would happen.

## Test instructions

* _N/A_
    As it's neigh impossible to fake server variables, this can not be tested manually. A unit test _could_ be added for this. but so far, this whole class is untested.


johannadevos:
CR 👍

PR #519, opened 2019-07-18
Exclude from news sitemap schema fallback

igorschoester:
## Summary

* Fixes a bug where the `@type` in the schema output would be `NewsArticle` instead of `Article` for articles which are excluded from the news sitemap.

* Moved the term exclusion functions to keep the code DRY.
* Added a protected function around `get_post` to be able to mock it in the tests.

## Test instructions

* Go to the news settings and make sure you have a couple of post types turned on.
* Create a new post (or any/all of the activated above) and publish it.
* Go to the frontend of that post and check the schema output (tip: [GSDTT](https://search.google.com/structured-data/testing-tool)); it should have the `@type` of `NewsArticle`.
* Go to your news sitemap, the newly created post should be in there too.

### Exclude from news sitemap
1. Now turn on the `Exclude from News Sitemap` option in the `Addons` tab -> `Google News` and save the post.
1. Check the schema and sitemap again: it should have the `@type` `Article` and not be in the sitemap.
1. Turn the `Exclude from News Sitemap` back off.

### Exclude term
1. Add a category to the post.
1. Go to your news settings and exclude the category from the news sitemap.
1. Check the schema and sitemap again: it should have the `@type` `Article` and not be in the sitemap.
1. Repeat the steps with a tag instead of a category.

### Other posttypes
* Repeat all the steps with another post type.

Fixes https://github.com/Yoast/bugreports/issues/326

### DIFF
--- a/classes/schema.php
+++ b/classes/schema.php
@@ -28,8 +28,8 @@ class WPSEO_News_Schema {
 	 * @return array $post_types Supported post types.
 	 */
 	public function article_post_types( $post_types ) {
-		// When the news article is excluded do not alter the article post types.
-		if ( $this->is_post_excluded() ) {
+		// Alter the article post types only when the news article is not excluded.
+		if ( ! $this->is_post_excluded() ) {
 			$post_types = array_merge( WPSEO_News::get_included_post_types(), $post_types );
 		}
 
@@ -46,8 +46,8 @@ class WPSEO_News_Schema {
 	public function change_article( $data ) {
 		$post = $this->get_post();
 		if ( $post !== null && in_array( $post->post_type, WPSEO_News::get_included_post_types(), true ) ) {
-			// When the news article is excluded do not change the `@type` to `NewsArticle`.
-			if ( $this->is_post_excluded( $post ) ) {
+			// Change the `@type` to `NewsArticle` only when the news article is not excluded.
+			if ( ! $this->is_post_excluded( $post ) ) {
 				$data['@type'] = 'NewsArticle';
 			}
 			$data['copyrightYear']   = mysql2date( 'Y', $post->post_date_gmt, false );
@@ -70,9 +70,9 @@ class WPSEO_News_Schema {
 		}
 
 		return (
-			$post !== null
-			&& WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false
-			&& WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
+			$post === null
+			|| WPSEO_News::is_excluded_through_sitemap( $post->ID )
+			|| WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type )
 		);
 	}
 
--- a/classes/schema.php
+++ b/classes/schema.php
@@ -70,16 +70,16 @@ class WPSEO_News_Schema {
 		}
 
 		return (
-			$post !== null &&
-			WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false &&
-			WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
+			$post !== null
+			&& WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false
+			&& WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
 		);
 	}
 
 	/**
 	 * Retrieves post data given a post ID or post object.
 	 *
-	 * This function exists to be abe to mock the get_post call and should
+	 * This function exists to be able to mock the get_post call and should
 	 * no longer be needed when moving the tests suite over to BrainMonkey.
 	 *
 	 * @codeCoverageIgnore
--- a/classes/schema.php
+++ b/classes/schema.php
@@ -55,11 +55,14 @@ class WPSEO_News_Schema {
 	/**
 	 * Retrieves post data given a post ID or post object.
 	 *
+	 * This function exists to be abe to mock the get_post call and should
+	 * no longer be needed when moving the tests suite over to BrainMonkey.
+	 *
 	 * @codeCoverageIgnore
 	 *
-	 * @param int|WP_Post|null $post Optional. Post ID or post object. Defaults to global $post.
+	 * @param int|WP_Post|null $post Optional. Post ID or post object.
 	 *
-	 * @return WP_Post|null Type corresponding to $output on success or null on failure.
+	 * @return WP_Post|null The post object or null if it cannot be found.
 	 */
 	protected function get_post( $post = null ) {
 		return get_post( $post );
--- a/classes/schema.php
+++ b/classes/schema.php
@@ -11,7 +11,7 @@
 class WPSEO_News_Schema {
 
 	/**
-	 * WPSEO_News_Head Constructor.
+	 * WPSEO_News_Schema Constructor.
 	 */
 	public function __construct() {
 		add_filter( 'wpseo_schema_article_post_types', array( $this, 'article_post_types' ) );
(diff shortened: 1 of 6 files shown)


tests/schema-test.php
```diff
@@ -0,0 +1,120 @@
...
+				'@id' => 'http://example.org#organization',
+			),
+		);
+		$actual   = $this->default_mock->change_article( array() );
+
+		$this->assertEquals( $expected, $actual );
+	}
+
+	/**
+	 * Tests whether the admin page is generated correctly with an excluded term.
```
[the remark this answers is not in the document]
IreneStr:
```suggestion
	 * Tests whether the schema output is generated correctly if one of the terms that is attached to a post is excluded from the news sitemap.
```

tests/schema-test.php
```diff
@@ -0,0 +1,120 @@
...
+				'@id' => 'http://example.org#organization',
+			),
+		);
+		$actual   = $this->default_mock->change_article( array() );
+
+		$this->assertEquals( $expected, $actual );
+	}
+
+	/**
+	 * Adds the options to exclude uncategorized posts. This is a filter function.
```
IreneStr: Moving this up to directly after setUp would make more sense to me.

tests/doubles/schema-double.php
```diff
@@ -0,0 +1,23 @@
...
+ * Class WPSEO_News_Schema.
+ */
+class WPSEO_News_Schema_Double extends WPSEO_News_Schema {
+
+	/**
+	 * Retrieves post data given a post ID or post object.
+	 *
+	 * @param int|WP_Post|null $post Optional. Post ID or post object. Defaults to global $post.
+	 *
+	 * @return WP_Post|array|null Type corresponding to $output on success or null on failure.
```
IreneStr:
```suggestion
	 * @return WP_Post|null The post object or null if it cannot be found.
```

tests/doubles/schema-double.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Class WPSEO_News_Schema.
+ */
+class WPSEO_News_Schema_Double extends WPSEO_News_Schema {
+
+	/**
+	 * Retrieves post data given a post ID or post object.
+	 *
+	 * @param int|WP_Post|null $post Optional. Post ID or post object. Defaults to global $post.
```
IreneStr:
```suggestion
	 * @param int|WP_Post|null $post Optional. Post ID or post object.
```

IreneStr:
Acceptance progress:
- Everything is working well for posts. 

To do: 
- Custom post types + page with and without excluded categories and tags.

tests/schema-test.php
igorschoester: I usually put helpers on the bottom of the file. We do not have any rules for this though. Because it is a class it will be loaded in before instantiated, so order does not matter for that reason.

classes/schema.php
```diff
@@ -39,13 +44,51 @@ public function article_post_types( $post_types ) {
...
 	public function change_article( $data ) {
-		$post = get_post();
-		if ( in_array( $post->post_type, WPSEO_News::get_included_post_types(), true ) ) {
-			$data['@type']           = 'NewsArticle';
+		$post = $this->get_post();
+		if ( $post !== null && in_array( $post->post_type, WPSEO_News::get_included_post_types(), true ) ) {
+			// When the news article is excluded do not change the `@type` to `NewsArticle`.
+			if ( $this->is_post_excluded( $post ) ) {
+				$data['@type'] = 'NewsArticle';
+			}
```
afercia: Not sure this is intended, or maybe the comment is misleading? I read "do not change the `@type` to `NewsArticle`." but then it does change the @type to `NewsArticle`?

classes/schema.php
```diff
@@ -39,13 +44,51 @@ public function article_post_types( $post_types ) {
...
+			$post !== null &&
+			WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false &&
+			WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
+		);
+	}
+
+	/**
+	 * Retrieves post data given a post ID or post object.
+	 *
+	 * This function exists to be abe to mock the get_post call and should
```
afercia: Typo: abe -> able

classes/wpseo-news.php
```diff
@@ -91,7 +91,7 @@ private function init_admin() {
 
 		// Setting action for removing the transient on update options.
 		if ( class_exists( 'WPSEO_Sitemaps_Cache' )
-			&& method_exists( 'WPSEO_Sitemaps_Cache', 'register_clear_on_option_update' )
+			 && method_exists( 'WPSEO_Sitemaps_Cache', 'register_clear_on_option_update' )
```
afercia: Why this added space?

classes/schema.php
```diff
@@ -39,13 +44,51 @@ public function article_post_types( $post_types ) {
...
+	 */
+	private function is_post_excluded( $post = null ) {
+		if ( $post === null ) {
+			$post = $this->get_post();
+		}
+
+		return (
+			$post !== null &&
+			WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false &&
+			WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
```
afercia: I guess for consistency with what we do in Yoast SEO, the `&&` should go at the beginning of the new lines.

classes/wpseo-news.php
```diff
@@ -162,8 +162,8 @@ public function plugin_links( $links, $file ) {
 		}
 		if ( $file === $this_plugin ) {
 			$settings_link = '<a href="' . admin_url( 'admin.php?page=wpseo_news' ) . '">'
-				. __( 'Settings', 'wordpress-seo-news' )
-				. '</a>';
+							 . __( 'Settings', 'wordpress-seo-news' )
+							 . '</a>';
 			array_unshift( $links, $settings_link );
```
afercia: Instead of this kind off indentation (which personally I think should be avoided because they use mixed tab characters and spaces) maybe a `sprintf()` would be cleaner.

classes/wpseo-news.php
```diff
@@ -162,8 +162,8 @@ public function plugin_links( $links, $file ) {
 		}
 		if ( $file === $this_plugin ) {
 			$settings_link = '<a href="' . admin_url( 'admin.php?page=wpseo_news' ) . '">'
-				. __( 'Settings', 'wordpress-seo-news' )
-				. '</a>';
+							 . __( 'Settings', 'wordpress-seo-news' )
+							 . '</a>';
```
afercia: Instead of this kind off indentation (which personally I think should be avoided because they use mixed tab characters and spaces) maybe a `sprintf()` would be cleaner.

classes/wpseo-news.php
```diff
@@ -261,7 +261,7 @@ public function error_missing_wpseo() {
 	public function error_upgrade_wp() {
 		echo '<div class="error"><p>';
 		printf(
-			/* translators: %1$s resolves to News SEO */
+		/* translators: %1$s resolves to News SEO */
```
afercia: this translators comment should go inside `esc_html__()`: must be on the line immediately preceding the string with placeholders.

classes/wpseo-news.php
```diff
@@ -279,7 +279,7 @@ public function error_upgrade_wp() {
 	public function error_upgrade_wpseo() {
 		echo '<div class="error"><p>';
 		printf(
-			/* translators: %1$s resolves to Yoast SEO, %2$s resolves to News SEO */
+		/* translators: %1$s resolves to Yoast SEO, %2$s resolves to News SEO */
```
afercia: same as above

classes/wpseo-news.php
```diff
@@ -240,7 +240,7 @@ public function enqueue_admin_page() {
 	public function error_missing_wpseo() {
 		echo '<div class="error"><p>';
 		printf(
-			/* translators: %1$s resolves to the link to search for Yoast SEO, %2$s resolves to the closing tag for this link, %3$s resolves to Yoast SEO, %4$s resolves to News SEO */
+		/* translators: %1$s resolves to the link to search for Yoast SEO, %2$s resolves to the closing tag for this link, %3$s resolves to Yoast SEO, %4$s resolves to News SEO */
```
afercia: this translators comment should go inside esc_html__(): must be on the line immediately preceding the string with placeholders.

classes/schema.php
afercia: I checked the actual behavior and though it works, I think part of the confusion here is that (_If I'm not wrong_) `is_post_excluded()` returns true when the post is NOT excluded and false when it is excluded. I'd say either it should be renamed or the various conditions swapped.

classes/schema.php
```diff
@@ -39,13 +44,51 @@ public function article_post_types( $post_types ) {
...
+	 */
+	private function is_post_excluded( $post = null ) {
+		if ( $post === null ) {
+			$post = $this->get_post();
+		}
+
+		return (
+			$post !== null
+			&& WPSEO_News::is_excluded_through_sitemap( $post->ID ) === false
+			&& WPSEO_News::is_excluded_through_terms( $post->ID, $post->post_type ) === false
```
afercia: Checking for `=== false` make this method actually return `true` when the post is NOT excluded. Maybe some leftover. See also previous comment https://github.com/Yoast/wpseo-news/pull/519/files#r307685526

classes/schema.php
```diff
@@ -26,7 +28,10 @@ public function __construct() {
 	 * @return array $post_types Supported post types.
 	 */
 	public function article_post_types( $post_types ) {
-		$post_types = array_merge( WPSEO_News::get_included_post_types(), $post_types );
+		// When the news article is excluded do not alter the article post types.
+		if ( $this->is_post_excluded() ) {
+			$post_types = array_merge( WPSEO_News::get_included_post_types(), $post_types );
```
afercia: `array_merge` usage is pre-existing but just wanted to note that this will produce something like the following, where the item `post` is present twice:

`array(3) { [0]=> string(4) "post" [1]=> string(4) "page" [2]=> string(4) "post" }`

> if the arrays contain numeric keys, the later value will not overwrite the original value, but will be appended.

classes/schema.php
IreneStr: I've reversed the logic to let it make more sense.

Xyfi:
I'm not sure whether this is in the scop of this issue but if a category is excluded from the sitemap the type is still `NewsArticle`.

PR #529, opened 2019-08-06
QA: use gmdate() instead of date()

jrfnl:
## Summary

* _N/A_

> WordPress core sets PHP time zone to UTC during boot. However it cannot guarantee that timezone isn't changed by third party code, which is relatively common in the wild.
>
> date() calls rely on current PHP time zone. If time zone is changed from UTC it causes invalid results through the core and extensions code.

Refs:
* https://core.trac.wordpress.org/ticket/46438
* https://github.com/WordPress/WordPress-Coding-Standards/issues/1713

Note: WPCS 2.2.0 (not yet released) will contain a check for the use of `date()`.

## Test instructions

* _N/A_ These changes only affect unit test files, so as long as those still pass, we should be good.


jrfnl:
Rebased for merge conflicts.

johannadevos:
CR & acceptance: 👍

PR #530, opened 2019-08-07
Don't copy files we don't need in the zip

IreneStr:
## Summary

* Makes sure we don't copy files we don't need in the zip. 

* When I compared the zip that was build before https://github.com/Yoast/Wiki/pull/90, and the one with https://github.com/Yoast/Wiki/pull/90, I noticed some differences. This PR removes the files that were never included in the zip, but would otherwise be included after https://github.com/Yoast/Wiki/pull/90.

## Test instructions

* See https://github.com/Yoast/Wiki/pull/90

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -8,6 +8,13 @@ module.exports = {
 					// Folders to copy.
 					"vendor/**",
 					"!vendor/bin",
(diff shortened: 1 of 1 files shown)


afercia:
CR and Acceptance 👍

PR #123, opened 2015-03-12
Create fallback wrappers for backwards compatibility

andizer:

PR #540, opened 2019-09-20
Deprecation of the wrappers

andizer:
## Summary

*  N/A - Cleanup of the wrappers

* I decided to deprecate wrappers because they are introduced (#123) back in 2015 for backwards compatibility reasons. We should call the `Yoast_Form` directly instead of fallbacks for methods that are deprecated since WPSEO 2.0 which also are removed in 2018 (https://github.com/Yoast/wordpress-seo/pull/8874). 

## Test instructions

* Make sure the News SEO page still works by editing fields and verifying values are still saved.

Fixes #


igorschoester:
CR & Acceptance ✅

PR #542, opened 2019-09-26
 PHPCompatibility fix: pass encoding to htmlspecialchars() et al functions

jrfnl:
## Summary

* _N/A_

(well, unless anyone ever reported a bug about this, but chances of that are small as it would only really come into play when using UTF-8 characters in the images titles and captions in combination with PHP < 5.4)

From the PHP 5.4 changelogs:
> The default character set for htmlspecialchars() and htmlentities() is now UTF-8, instead of ISO-8859-1.

While _most_ of the WP world uses the WP default `utf-8` character encoding, this can be changed, so to be on the safe side, one should always pass the encoding to functions affected by it.

Refs:
* https://www.php.net/manual/en/migration54.other.php
* https://www.php.net/manual/en/function.htmlentities.php#refsect1-function.htmlentities-changelog
* https://www.php.net/manual/en/function.htmlspecialchars.php#refsect1-function.htmlspecialchars-changelog

## Test instructions

**_Only for die-hard testers_**

On `trunk`:
* Make sure you have both PHP 5.2/5.3 and a higher PHP version installed.
* Make sure the WP install is on UTF-8.
* Add a news item with an image with a title and a caption using UTF-8 characters - think something in Russian or Chinese would be good.
* View the news image sitemap on both PHP versions.
    In PHP 5.2/5.3, the texts will be "weird", i.e. contain wrongly encoded characters. In the higher PHP version all should be fine.

Rinse & repeat with this branch and all should be good on both PHP versions.

### DIFF
--- a/classes/sitemap-images.php
+++ b/classes/sitemap-images.php
@@ -175,15 +175,16 @@ class WPSEO_News_Sitemap_Images {
 		 */
 		$src = apply_filters( 'wpseo_xml_sitemap_img_src', $src, $this->item );
 
+		$encoding      = get_bloginfo( 'charset' );
 		$this->output .= "\t<image:image>\n";
-		$this->output .= "\t\t<image:loc>" . htmlspecialchars( $src ) . "</image:loc>\n";
+		$this->output .= "\t\t<image:loc>" . htmlspecialchars( $src, ENT_COMPAT, $encoding, false ) . "</image:loc>\n";
 
 		if ( ! empty( $img['title'] ) ) {
-			$this->output .= "\t\t<image:title>" . htmlspecialchars( $img['title'] ) . "</image:title>\n";
+			$this->output .= "\t\t<image:title>" . htmlspecialchars( $img['title'], ENT_COMPAT, $encoding, false ) . "</image:title>\n";
 		}
 
 		if ( ! empty( $img['alt'] ) ) {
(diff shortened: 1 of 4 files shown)


andizer:
CR done 👍 

With that specific charset set in WordPress I would expect there is a WordPress equivalent for it like `wp_htmlspecialchars`. Is it something that is useful to have in WordPress or it is too minor?

jrfnl:
> With that specific charset set in WordPress I would expect there is a WordPress equivalent for it like wp_htmlspecialchars. Is it something that is useful to have in WordPress or it is too minor?

Don't get me started on consistent charset handling with WP Core.... whole different story and sh*tloads which needs to be improved. :scream:
Luckily with the new minimum PHP version of 5.6, the PHP `Intl` extension is now available, so that could be used to make those improvement if anyone dares to take up the mantle...

Re: `wp_htmlspecialchars()` - I'm not aware of a WP function for this (doesn't mean it doesn't exist), but it not that relevant for Core now the minimum PHP version is 5.6, which defaults to `utf-8`.

For a number of functions the default changed in PHP 5.4, for `iconv` and `mbstring`, the default changed in PHP 5.6, so the problem in this case is largely superseded by the new minimum PHP version, though passing `$encoding` should still always be the recommended best practice.

Not super helpful as an answer, but hopefully gives you enough context ? :grimacing:

PR #544, opened 2019-10-02
Travis: minor clean up

jrfnl:
## Summary
* _N/A_

Remove redundant, duplicate `composer install`.

The `script` will use the Travis provided PHPUnit version when compatible and use the Composer installed one when not.
See: https://github.com/Yoast/wpseo-news/blob/af1ebddf58a4f568e46e9258549fbcc1609cc7eb/.travis.yml#L176-L197

The preceding `composer install` already takes this into account.
See: https://github.com/Yoast/wpseo-news/blob/af1ebddf58a4f568e46e9258549fbcc1609cc7eb/.travis.yml#L82-L88

So, there is no need to run a second `composer install` based on the unit test ENV variables.

## Test instructions

* _N/A_
    This is a build-script-only change and should have no effect on the functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,10 +87,6 @@ install:
     composer install --no-dev --no-interaction
   fi
 - |
-  if [[ "$PHPUNIT" == "1" || "$PHPUNIT_INTEGRATION" == "1" ]]; then
-    composer install --no-interaction
-  fi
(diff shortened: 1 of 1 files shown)


IreneStr:
CR 👍

PR #546, opened 2019-10-17
Rename new testcase file

jrfnl:
## Summary
* _N/A_

Rename new testcase file to comply with the CS rules.

## Test instructions

* _N/A_
    This is a filename-only change and should have no effect on the functionality.


johannadevos:
CR: 👍

PR #547, opened 2019-10-17
CS/QA: improve output escaping

jrfnl:
## Summary
* _N/A_

Post type and taxonomy labels should never contain HTML, so can safely be escaped with `esc_html()`.

To that end, move the `esc_html()` from inside the `sprintf()` to wrapping the `sprintf()`.

## Test instructions

* _N/A_
    This is a code-only change and should have no effect on the functionality. Well, that is, unless someone would happen to add HTML to the post type/taxonomy labels, in that case, the display will now be different.


andizer:
CR done

PR #553, opened 2019-10-29
Travis: fix broken script

jrfnl:
## Summary
This PR can be summarized in the following changelog entry:
* _N/A_

## Relevant technical choices:

The Travis builds have been failing for quite a long time, but due to the script being set up incorrectly, this has gone (nearly) unnoticed.

The main problems were:
* Incorrect indentation.
* Incorrect folding.
* Commands where the exit code should determine build pass/fail being nested.
    Travis does not handle nested conditions well.

The above has all been cleaned up and the build is now correctly failing on all builds which run unit tests.

In addition to this:
* Removing `sudo`.
    Travis has not supported `sudo` for quite a while.
* Moving the merging of the code coverage test output to the `after_script` section.

**IMPORTANT NOTE**:
**The build for this PR will not pass until all the failing unit tests have been fixed and this branch has been rebased after merging the unit test fixes.**

### Failures which were "hidden" until now

These are screenshots of the current failures as per the latest build against `trunk`.

https://travis-ci.org/Yoast/wpseo-news/jobs/604592951#L664-L697 / PHP 7.2 / COVERAGE=1

image

https://travis-ci.org/Yoast/wpseo-news/jobs/604592953#L330-L331 / PHP 5.2 / PHPUNIT_INTEGRATION=1

image

PRs to fix these failures are upcoming.

## Test instructions

This PR can be tested by following these steps:
* Check the **DETAILED** output of the Travis script, not just the pass/fail return value.

Fixes #402

PR #554, opened 2019-10-29
Schema: fix failing test

jrfnl:
## Summary
* _N/A_

This test has been failing for a while - since commit 460da5aa4903f831ca3ab0659708516ae48ba594 - and the failure was hidden by errors in the Travis script.

The change made in the above mentioned commit means that the URL for `copyrightHolder` will always have a trailing slash before the hash `#`.
The unit tests were not updated for this change, causing them to fail.

## Test instructions

* Run the integration tests locally against a high PHP version in combination with `trunk` and see them fail.
* Switch to this branch and see the unit tests passing.


igorschoester:
LGTM 👍

PR #558, opened 2019-10-29
Fix more failing tests

jrfnl:
## Summary
* _N/A_

These tests have been failing for a while (if not forever) and the failure was hidden by errors in the Travis script.

```
Fatal error: Call to undefined method Mock_WPSEO_News_Meta_Box_Double_d7df196f::method() in /tmp/wordpress/src/wp-content/plugins/wpseo-news/integration-tests/meta-box-test.php on line 32
```

The short of it is that these tests are using more modern PHPUnit syntax, while the tests are also still being run on older PHPUnit versions where that syntax was not yet available.

## Test instructions

* Run the integration tests locally against `trunk` in combination with WP 5.1 on PHP 5.2 with PHPUnit 3.x and see them fail.
* Run them again against this branch and see a different error, but no longer the one mentioned above.

### DIFF
--- a/integration-tests/meta-box-test.php
+++ b/integration-tests/meta-box-test.php
@@ -29,9 +29,17 @@ class WPSEO_News_Meta_Box_Test extends WPSEO_News_UnitTestCase {
 			)
 			->getMock();
 
-		$stub->method( 'is_post_type_supported' )->willReturn( true );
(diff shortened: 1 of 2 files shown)


jrfnl:
As the PHP 5.2 drop is happening slightly earlier than anticipated, this fix is no longer necessary, though the script fix in PR #553 won't pass without it.

PR #560, opened 2019-11-02
Stop explicit support for PHP < 5.6 / WP < 5.2

jrfnl:
## Summary
* _N/A_

> ~~This PR requires PR #553 to be merged first~~
> ~~Once that PR has been merged, I will rebase this PR and change it from Draft to Mergable.~~

Brief analysis of the Travis script shows me that:
* There are a number of builds _which do not do anything_.
    All the tasks in `script` are set up conditionally based on environment variables.
    - Any build which did not have an environment variable, such as `PHPLINT`, `PHPCS` or `PHPUNIT` is basically completely useless, i.e the `PHP 7.3 WP_VERSION=master` build.
* There are three different environment variables related to toggles for running the unit tests:
    - `PHPUNIT` - to run the new BrainMonkey based tests.
    - `PHPUNIT_INTEGRATION` - to run the old integration tests.
    - `COVERAGE` - to run both and generate code coverage.
    While the basic setup for the BrainMonkey tests has been done, there are currently no such tests for the News plugin.
	So basically any build which only has the `PHPUNIT` environment variable for testing is - again - useless, i.e. the `PHP 5.6` and the  `PHP 7.4` builds.
* While the unit test related tasks are set up to run conditionally, the _preparation of the environment_ for running the integration tests, i.e. downloading WordPress, cloning YoastSEO Free, setting up the database _is all done **unconditionally**_.

So, with that in mind, I have made the following changes to the Travis script - please read this carefully:
* Removing builds against PHP < 5.6.
* Changing the `WP_VERSION` environment variables in the matrix to use a minimum of `5.2`.
    Note: this will change again once WP 5.3 comes out.
* Actually running the unit tests on all builds.
* Removing the `PHPUNIT_INTEGRATION` environment variable and setting up the `PHPUNIT` environment variable to run both the integration tests as well as the BrainMonkey tests.
* Removing the toggle for a full/no-dev composer install.
    As a composer install was being run on all builds anyway in one way or another **and** there is a committed `composer.lock` file **and** Travis caching of the downloads is set up, doing a full install each time will not make the build significantly slower.
    Also: It is better to always test based on the dependencies we've set up than to rely on a PHP image created by Travis which may fail unexpectedly (especially as Travis has seen some issues recently with incorrect PHPUnit versions on certain images).
    This also means the `composer dump-autoload` can be removed as it will now always have run whenever the autoload is needed.
    **Note**: this changes the `composer install` for the `deploy` stage from a full install to a `no-dev` install.
   That stage was currently being run against PHP 7.2, so was getting a full install, while AFAICS that shouldn't be needed.
* Adding conditions around the commands setting up the integration test environment.
    Downloading WP and YoastSEO is only needed when the unit tests are run.
* Using the composer scripts to run various tasks.
    As PHPUnit will now always be available in `vendor`, using the composer scripts simplifies things.

Includes:
* Removing the `xrstf/composer-php52` dependency and references to the `vendor/autoload_52.php` file.
* Dropping PHPUnit < 5.7 from the Composer `require-dev`.

Related to https://github.com/Yoast/wordpress-seo/issues/13758

## Test instructions

* Check the detailed output of the Travis scripts to verify that all is working as expected.

**_Please be extra critical reviewing this PR for any changes which may impact the `deploy` stage as those won't show up during the PR build._**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,19 +26,15 @@ jobs:
   fast_finish: true
   include:
     - php: 7.3
-      env: WP_VERSION=master
+      env: WP_VERSION=master PHPUNIT=1
     - php: 7.2
-      env: WP_VERSION=latest WP_MULTISITE=1 PHPLINT=1 PHPCS=1 SECURITY=1 COVERAGE=1
-    - php: 5.2
-      # As 'trusty' is not supporting PHP 5.2/5.3 anymore, we need to force using 'precise'.
-      dist: precise
-      env: WP_VERSION=5.0 WP_MULTISITE=1 PHPLINT=1 PHPUNIT_INTEGRATION=1
+      env: WP_VERSION=5.2 WP_MULTISITE=1 PHPLINT=1 PHPCS=1 SECURITY=1 COVERAGE=1
     - php: 5.6
-      env: WP_VERSION=latest PHPUNIT=1
+      env: WP_VERSION=5.2 WP_MULTISITE=1 PHPLINT=1 PHPUNIT=1
       # Use 'trusty' to test against MySQL 5.6, 'xenial' contains 5.7 by default.
       dist: trusty
     - php: "7.4snapshot"
-      env: WP_VERSION=latest PHPUNIT=1
+      env: WP_VERSION=master PHPUNIT=1
     - stage: deploy-to-github-dist
       env: WP_VERSION=latest
       if: tag IS present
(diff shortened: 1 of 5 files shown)


igorschoester:
This can be rebased now 😄

jrfnl:
@igorschoester Thanks for the ping. Done.

jrfnl:
@andizer asked a good question about the change regarding the deploy stage check in a similar PR.
My reply to that question may also be useful to read through for this PR: https://github.com/Yoast/wpseo-woocommerce/pull/433#issuecomment-551241420

PR #569, opened 2019-11-18
Travis: don't allow builds against PHP 7.4 to fail

jrfnl:
## Summary
* _N/A_

Includes:
* Adjusting the Travis matrix now WP 5.3 has been released
* Removing the build against PHP 7.4 from `allowed failures`.
* Adding (back) a build against PHP `nightly` (PHP 8) and adding it to  `allowed_failures`.
    For now, this build will only lint the code. Unit testing is not yet possible with  the current setup.

## Test instructions

* Check the detailed output of the Travis script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,13 +28,15 @@ jobs:
     - php: 7.3
       env: WP_VERSION=master PHPUNIT=1
     - php: 7.2
-      env: WP_VERSION=5.2 WP_MULTISITE=1 PHPLINT=1 PHPCS=1 SECURITY=1 COVERAGE=1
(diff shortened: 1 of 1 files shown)


igorschoester:
Why is the PHP 7.3 build still allowed to fail?

Besides that, good to merge!

jrfnl:
> 
> 
> Why is the PHP 7.3 build still allowed to fail?

That is _only_  PHP 7.3 in combination with WP `master`, which is basically `trunk`, which is WP 5.5.

So it's not about PHP 7.3 being allowed to fail, but about the build against the next version of WP, which is currently still unstable, being allowed to fail.

Does that answer your question?

igorschoester:
Yes, thank you 😄 
Merging!

PR #578, opened 2019-11-28
Add missing multisite checks for the actions

igorschoester:
## Summary

* Fixes a bug where the sitemap could be invalidated twice on multisite with MultilingualPress.

*

## Test instructions

These are kind of hard to test. The idea is that the code could get called twice when on multisite.
Therefore, some debugging is needed to verify this. [See this Wiki page to learn how](https://github.com/Yoast/Wiki/wiki/PHPStorm-debugging).

*Note: To acquire a copy of MultilingualPress, login to LastPass -> Shared-licenses vault. There's a login there for MultilingualPress so you can download the latest version.*

These are kind of hard to test. The idea is that the code could get called twice when on multisite.
Therefore, some debugging is needed to verify this. [See this Wiki page to learn how](https://github.com/Yoast/Wiki/wiki/PHPStorm-debugging).

*Note: To acquire a copy of MultilingualPress, login to LastPass -> Shared-licenses vault. There's a login there for MultilingualPress so you can download the latest version.*

Setup your multisite with MLP:
* Ensure you have a multisite environment with MultilingualPress and Yoast SEO installed.
* Configure at least two languages in MultilingualPress.

Test the `core` > `invalidate_sitemap`:
* Set a breakpoint in `classes/class-core.php` in the method `invalidate_sitemap` on lines 265 (the new `return`) and 268 (the first statement after the new `return`).
* Create a new post and ensure you set it to sync with the secondary language in the MLP metabox) and publish it. Please ignore the first debug call when you just clicked on create new post. This is the creation of the draft.
* You should see both debug lines executed now. The actual post (not the MLP match) is called twice (probably the revision).

Related https://github.com/Yoast/wordpress-seo/issues/13765

### DIFF
--- a/classes/sitemap.php
+++ b/classes/sitemap.php
@@ -98,6 +98,11 @@ class WPSEO_News_Sitemap {
 			return;
 		}
 
+		// Bail if this is a multisite installation and the site has been switched.
+		if ( is_multisite() && ms_is_switched() ) {
+			return;
+		}
+
 		// Only invalidate when we are in a News Post Type object.
 		if ( ! in_array( get_post_type( $post_id ), WPSEO_News::get_included_post_types(), true ) ) {
 			return;


afercia:
CR 👍

PR #586, opened 2019-12-12
QA: sanitize use of a superglobal

jrfnl:
## Summary
* _N/A_

* WP slashes all superglobals, so before actually using a index, it should be unslashed.
* Before doing a compare, sanitize the text string to clean it off null bytes and invalid UTF-8.

## Test instructions

* ... good question....


karlijnbok:
CR: 👍 
Acceptance: 👍 (testing was done by comparing the news sitemaps from trunk and this branch > no differences were found)

Issue #585, opened 2019-12-12
QA: Review use of a direct DB query without caching


jrfnl:
There is one query used within News which:
a) makes a direct database calls instead of using any of the various DB call abstraction functions offered by WP.
b) is not caching the results

This query should be carefully reviewed and where appropriate, fixed.
https://github.com/Yoast/wpseo-news/blob/5136c846e341268427146e74ed2feee561192a2b/classes/sitemap.php#L231-L243

This query is a `SELECT` query, so caching would seem appropriate.

Recommended for the refactoring:
* https://developer.wordpress.org/reference/classes/wp_meta_query/
* https://developer.wordpress.org/reference/classes/wp_query/

If after careful reviewing it is determined for this query that it is really REALLY ok the way it is and should not be cached or changed to use a WP DB abstractions, the query can be whitelisted using the following flags:
* `// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery`
* `// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching`
* or for both: `// phpcs:ignore WordPress.DB.DirectDatabaseQuery`

### Issue details
```
FILE: classes\sitemap.php                                                         
--------------------------------------------------------------------------------  
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 1 LINE                                    
--------------------------------------------------------------------------------  
 238 | WARNING | Usage of a direct database call is discouraged.                  
 238 | WARNING | Direct database call without caching detected. Consider using    
     |         | wp_cache_get() / wp_cache_set() or wp_cache_delete().            
--------------------------------------------------------------------------------  
```

PR #589, opened 2019-12-19
PHPCS: Update to YoastCS 1.3.0

jrfnl:
## Summary
* _N/A_

Note: As the News plugin was so far behind, the update to YoastCS 2.0 is done in two steps - first to YoastCS 1.3.0 (this PR), then to YoastCS 2.0.
This allows for slightly easier reviewing of the changes.

### Composer: up the requirement for YoastCS

* A selective update of the `composer.lock` for just YoastCS and its dependencies.
* The `installed_paths` command is now no longer needed. This will be handled automatically by the Composer plugin which comes with YoastCS 1.2.0+.
    To that end, remove the `composer config-yoastcs` command from the Travis script.
* To not break BC for devs who are used to having this command available, update the composer script command to call the Dealerdirect plugin to handle this.

### Build/PHPCS: Rename PHPCS configuration file

Using a `.dist` extension allows for individual developers to overload the `phpcs.xml.dist` with a local `phpcs.xml` file. This file can include the `dist` file as a base ruleset using `<rule ref="./.phpcs.xml.dist"/>`.

To facilitate this, the typical file names for the overloaded PHPCS config file have been added to the `.gitignore` file.

### Build/PHPCS: set up the PHPCS ruleset properly

* Add the XSD.
* Add fencing and documentation.
* Provide all command line arguments.
* Remove the exclusion of the `tests` directory.
* Remove the exclusion of the dependency related directories. These are excluded by default since YoastCS 1.3.0.
* Remove sniff `exclude`s which have clearly been copy/pasted from the original WPSEO ruleset.
* Don't hide PHPCS warnings as a default.
* Use tab based indentation in the ruleset file in line with indentation elsewhere.

### Build/PHPCS: list the custom test classes

Several sniffs will make exceptions for code in unit test classes.
This is determined by checking whether these classes extend known unit test case classes.

This property provides those class names to the sniffs.

### Build/PHPCS: specify the text-domain which should be used

... to enable the check for this within the WP.I18n sniff.

Includes fixing the `Text domain` header in the main plugin file which was using an incorrect text domain.

### Build/PHPCS: add a few selective exclusions

... includes documentation on why these files should be excluded for these specific rules.

* Allow for overloading method just to change visibility in test situations.
* Allow non-prefixed globals in select test situations.
* Allow access to and overloading of superglobals in the unit test case file.
* Allow using `@inheritdoc` for methods in test double classes.

### Build/PHPCS: temporary tweak for short array format

WordPressCS will demand long arrays and forbid the use of short arrays as of WPCS 2.2.0.
See: https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

For YoastCS, however, a choice has been made to demand short arrays.

As support for PHP < 5.6 has been dropped, this can be enforced everywhere now and will be as of YoastCS 2.0.0 which is expected soon.

For now, just ignore use of short arrays.

### Build/PHPCS: add temporary exclusion

This need further review/attention and has a `technical-debt` ticket attached to it - see #585.

### PHPCS: whitelist a direct database query

... which is only used when a new version of the plugin is installed and not as a regular page query.

## Test instructions

* Run `composer install`
* Run `composer check-cs` & see a clean result.


IreneStr:
CR 👍

PR #600, opened 2020-01-13
Fixes the performance

### LINKED ISSUE #592: Implement the checkbox list element.
In https://github.com/Yoast/wordpress-seo/pull/14122 a checkbox list element has been introduced we should use this for rendering the lists on the admin page for news seo. We can do this for the inclusion for post types and for the selection of taxonomies that will be excluded.

For the both lists we are saving each checkbox state separately and we have to convert these to two options that are an array.

- [ ] Implement the form element on the admin page
- [ ] Create two new options: `news_sitemap_include_post_types` and `news_sitemap_exclude_term_taxonomies` - The last one can be a little bit tricky because of its name. 
- [ ] Use the new options in the code base.
- [ ] Convert the current options to the new ones
- [ ] Remove the current options

andizer:
## Summary

* Fixes a bug where pages are loaded slowly

* Followed everything mentioned in this comment https://github.com/Yoast/bugreports/issues/822#issuecomment-573637790, except for the test for `is_excluded_through_terms` because that is too complex with static dependencies.

## Test instructions

* Makes sure the upgrade routine has ran successful, please set that version constant `WPSEO_NEWS_VERSION` to 12.4.1-RC1
* Use the settings for news and set the post type to exclude and verify the post is excluded
* Do the same for the terms.
* Test the performance on a site with many tags. See https://github.com/Yoast/bugreports/issues/822

Fixes https://github.com/Yoast/bugreports/issues/822
Fixes #592

### DIFF
--- a/classes/option.php
+++ b/classes/option.php
@@ -23,9 +23,11 @@ class WPSEO_News_Option extends WPSEO_Option {
 	 * @var array
 	 */
 	protected $defaults = array(
-		'news_sitemap_name'          => '',
-		'news_sitemap_default_genre' => '',
-		'news_version'               => '0',
+		'news_sitemap_name'               => '',
+		'news_sitemap_default_genre'      => '',
+		'news_version'                    => '0',
+		'news_sitemap_include_post_types' => array(),
+		'news_sitemap_exclude_terms'      => array(),
 	);
 
 	/**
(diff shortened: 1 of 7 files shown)


classes/option.php
```diff
@@ -129,16 +83,17 @@ protected function validate_option( $dirty, $clean, $old ) {
...
-							unset( $clean[ $key ] );
+				case 'news_sitemap_include_post_types':
+				case 'news_sitemap_exclude_terms':
+					$clean[ $key ] = array();
+
+					if ( isset( $dirty[ $key ] ) && ( is_array( $dirty[ $key ] ) && $dirty[ $key ] !== array() ) ) {
+						foreach ( $dirty[ $key ] as $name => $posted_value ) {
+							if ( is_string( $name ) ) {
+								$clean[ $key ][ $name ] = 'on';
+							}
```
IreneStr: Don't we want to keep a default?

IreneStr:
Merging this so QA can start testing. But please answer my question about the default case, @andizer :)

PR #649, opened 2020-08-13
P1-125 Fix NewsArticle not working when the current article type is None

igorschoester:
## Context

* Pages have the default article type of `None` (and this can not be changed), but still have author support. They should be able to be changed to `NewsArticle`.

## Summary

* Fixes a bug where the schema article would not be output for `Pages` and other post types with author support that have their current article type set to `None`.

* Use `wpseo_schema_article_type` to change the type, because the `is_needed` checks the type. And the `wpseo_schema_article` won't be triggered if the type is `None`.
* When the type includes `None`, this is changed to `Article` to make it consistant with post types that are not set to `None`. E.g. the `@type` will end up as `[ "Article", "NewsArticle" ]`, and not just `"NewsArticle"`.

## Test instructions

### Pages
* In the News SEO options, include the post type `Pages` in the news sitemap. This triggers the "trying to output the Article schema".
* Visit a page's frontend.
  * Check if there is an article graph piece. Where there was not one without this PR.
  * The `@type` should be `[ "Article", "NewsArticle" ]`.
* In the News SEO options, do not include `Pages` in the news sitemap.
* Recheck the frontend for the page, it should no longer have the article schema.

### Custom post type without author support
* When you use the Yoast test helper, you can turn on Books. It does not have author support.
* Include `Books` in the news sitemap.
* Visit a book's frontend.
  * There should __not__ be an article graph piece because of the lack of author support.

### Custom post type with author support
* If you don't have one yet, I can recommend the `Custom Post Type UI` plugin.
* Include the custom post type in the news sitemap.
* Create a "post" and visit the frontend.
  * Check if there is an article graph piece.
  * The `@type` should be `[ "Article", "NewsArticle" ]`.
* Verify changing the article type works:
  * Change the article type in the schema tab to something other than `None`.
  * The `@type` should change to include the given type. No duplicates should be there (so changing to `Article` and `NewsArticle` should not result in a change).
* You can repeat the "custom post type" checks for posts too.

Fixes https://yoast.atlassian.net/browse/P1-125

### DIFF
--- a/classes/schema.php
+++ b/classes/schema.php
@@ -5,6 +5,8 @@
  * @package WPSEO_News
  */
 
+use Yoast\WP\SEO\Config\Schema_IDs;
+
 /**
  * Makes the require Schema changes.
  */
@@ -26,7 +28,8 @@ class WPSEO_News_Schema {
 		$this->date = new WPSEO_Date_Helper();
 
 		add_filter( 'wpseo_schema_article_post_types', [ $this, 'article_post_types' ] );
-		add_filter( 'wpseo_schema_article', [ $this, 'change_article' ] );
+		add_filter( 'wpseo_schema_article_type', [ $this, 'add_news_article_type' ] );
+		add_filter( 'wpseo_schema_article', [ $this, 'add_copyright_information' ] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


classes/schema.php
```diff
@@ -47,31 +50,71 @@ public function article_post_types( $post_types ) {
...
+			$type = [ $type ];
+		}
+
+		/*
+		 * Replace `None` with `Article` if included.
+		 * This is to keep it consistent with post types that already include an Article.
+		 */
+		$type = array_map(
+			function ( $value ) {
+				return ( $value === 'None' ) ? 'Article' : $value;
```
maartenleenders, line 80: I feel like this should be the responsibility of the wordpress-seo plugin: if anyone edits the type via the filter to add a type, and we've added `None` (or they have added `None`), we'll get an array with the added type and None. 

Here you prevent an issue properly, but we can't rely on everyone implementing it correctly.

What do you think?

maartenleenders: Had a call with @igorschoester and decided to fix it after 14.8 in a separate issue (to be created).

maartenleenders:
CR 👍

maartenleenders:
Acceptance 👍

PR #652, opened 2020-10-14
Introduce `composer cs` command

moorscode:
## Context

* Introduces the `composer cs` command; as also present in Yoast SEO

## Summary

* Tooling: introduces the `composer cs` command

* Adjusted CS configuration to handle the config folder; inspired on Yoast SEO
* Created a config directory
* Removed the `yoast-configcs` command; which is no longer needed

## Test instructions

* Run `composer install`
* Run `composer cs` and get a menu to pick which check you want to perform
* Perform any of the menu options and see them working as expected

Fixes #

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
             "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs"
         ],
         "check-staged-cs": [
-            "@check-cs-warnings --filter=GitStaged"
+            "@check-cs --filter=GitStaged"
         ],
         "check-branch-cs": [
             "Yoast\\WP\\News\\Composer\\Actions::check_branch_cs"
--- a/composer.json
+++ b/composer.json
@@ -36,19 +36,25 @@
     "autoload-dev": {
         "classmap": [
             "integration-tests/",
-            "tests/"
+            "tests/",
+            "config/"
         ]
     },
     "scripts": {
         "lint": [
             "@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . -e php --exclude vendor --exclude node_modules --exclude .git"
         ],
-        "config-yoastcs": [
-            "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run",
-            "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs --config-set default_standard Yoast"
+        "cs": [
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -36,19 +36,25 @@
...
-            "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs --config-set default_standard Yoast"
+        "cs": [
+            "Yoast\\WP\\News\\Composer\\Actions::check_coding_standards"
         ],
         "check-cs": [
-            "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs --runtime-set ignore_warnings_on_exit 1"
+            "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs"
+        ],
+        "check-staged-cs": [
+            "@check-cs-warnings --filter=GitStaged"
```
hansjovis, line 54: This command fails, since no `check-cs-warning` command is defined within this `composer.json` file.

In WordPress SEO we have two similar commands. One that shows warnings:
```json
"check-cs-warnings": [
	"@php ./vendor/squizlabs/php_codesniffer/bin/phpcs"
],
```
and one that does not:
```json
"check-cs": [
	"@check-cs-warnings -n"
],
```

Since I am not sure which one we want here (or if we want both), I am going to give this one back to you 😉.

config/composer/actions.php
```diff
@@ -0,0 +1,117 @@
...
+		\exec( 'git diff --name-only --diff-filter=d ' . \escapeshellarg( $compare ), $files );
+
+		$php_files = self::filter_files( $files, '.php' );
+		if ( empty( $php_files ) ) {
+			echo 'No files to compare! Exiting.' . PHP_EOL;
+
+			return 0;
+		}
+
+		\system( 'composer check-cs-warnings -- ' . \implode( ' ', \array_map( 'escapeshellarg', $php_files ) ), $exit_code );
```
hansjovis, line 97: This fails as well for the same reason: `check-cs-warnings` is not defined.

hansjovis (changes requested):
Reviewed and tested this. Some added Composer commands fail since `check-cs-warnings` is not defined (see the commment).

moorscode:
Applied code review comments.

johannadevos:
Acceptance: ✅ 
- [x] All 4 `composer cs` commands run without problems.

PR #656, opened 2020-11-14
Unit Tests: make the testsuite compatible with PHPUnit 9 and test against PHP 8

jrfnl:
## Context

* Preparation for compatibility with PHP 8

## Summary

* Preparation for compatibility with PHP 8

### Composer: require Yoast/wp-test-utils

* Adds a dev dependency to the `yoast/wp-test-utils` package, which will also  make the PHPUnit Polyfills available.
* As that package already requires and manages the installable versions for  BrainMonkey and PHPUnit, remove these as explicit requirements from `require- dev` in favour of letting the WP Test Utils package manage the versions.

The install is done based on PHP 5.6 and includes updating the dependencies  from PHPUnit itself within the constraints of maintaining compatibility with  PHP 5.6.

This new package adds the following features:
* Polyfills for various PHPUnit cross-version changes.
* Basic test cases for use by the plugins.

Refs:
* https://github.com/Yoast/wp-test-utils
* https://github.com/Yoast/PHPUnit-Polyfills/

### Tests: use the bootstrap file from WP Test Utils

... and remove everything which is already done within that file.

### Tests: switch over to the Yoast\WPTestUtils\BrainMonkey\YoastTestCase

All test classes for the BrainMonkey based unit tests in the NewsSEO test suite have a NewsSEO native `TestCase` as a parent class.

This switches the parent of that `TestCase` from the PHPUnit native TestCase to the `Yoast\WPTestUtils\BrainMonkey\YoastTestCase`.

The `YoastTestCase`:
* Inherits PHPUnit cross-version compatibility from the PHPUnit Polyfills TestCase.
* Inherits the BrainMonkey set up and teardown and Mockery expectation counting as assertions from the BrainMonkey TestCase.
* And natively adds a number of stubs for commonly used WP functions.

This switch over includes:
* Removing the addition of function stubs which are already added via the `YoastTestCase` class.
* Removing the `tearDown()` method. This will be inherited from the `YoastTestCase` now.
* Renaming the `setUp()` method to `set_up()` in various test classes for PHPUnit cross-version compatibility.

### TestCase: remove the generic stubbing of the translation and escaping functions

Turns out that none of the (current) tests need these stubs, so we may as well not load them.

### Travis: run the unit tests against PHP 8/nightly

Now the BrainMonkey based tests are fully compatible with PHPUnit 9, the test suite can be run on PHP 8.

However, as the integration tests are not (yet) compatible with PHPUnit > 5.7, we need to add a separate condition block to the `script` section.

In this block, we run a `composer update` for just and only the test dependencies on PHP 8 to make sure that the test dependencies available are compatible with PHP 8.
And as the `composer.json` contains a hard-coded `platform.php` setting, we use `ignore-platform-reqs`  to get round that setting.

Includes:
* Adding `travis_retry` to `composer install`-like commands to get round builds stalling on the connection with Packagist, especially on older PHP versions.
* Adding the PHPUnit cache file which is generated when using PHPUnit 8/9 to `.gitignore`.

### PHPUnit config: minor tweaks

* Improve the code coverage configuration:
    - Add `addUncoveredFilesFromWhitelist` and  `processUncoveredFilesFromWhitelist` directives with sensible values.
    - Add the `forceCoversAnnotation` attribute to only record code coverage for the method under test as annotated via a `@covers` annotation.

Based on this configuration, the strict code coverage for the BrainMonkey based unit testsuite alone is currently **7.11%**.

* Verify the output of all builds for this PR, making sure that 1) the installation via Composer works without issues, 2) the tests are being run and 3) the tests pass.

To test locally, it will be simplest to:
1. Run `composer install`.
2. Run `composer test` on any PHP version below 8.0 and see the tests pass, including a more accurate assertion count (48 compared to 10 previously, due to Mockery expectations now being counted).
    This confirms that the changes to the `TestCase` works correctly cross-version as the PHPUnit version installed based on the `composer.lock` file will be PHPUnit 5.7.27.
3. Switch to PHP 8.
4. Run `composer update yoast/wp-test-utils --with-dependencies --ignore-platform-reqs`
5. Run `composer test` to see the tests running and passing on PHP 8 in combination with PHPUnit 9.x.
    This confirms that the tests can now run cross-version on all PHPUnit versions between PHPUnit 5.7 and PHPUnit 9.x and that the tests run and pass on PHP 8.

Note: this doesn't confirm full PHP 8 compatibility yet as the test code coverage base on the BrainMonkey based tests alone, as noted above, is only ~7%.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ phpunit.xml
 phpcs.xml
 .phpcs.xml
 .cache/phpcs.cache
+.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
     - php: 7.4
       env: WP_VERSION=latest PHPLINT=1 PHPUNIT=1
     - php: "nightly"
-      env:  PHPLINT=1
+      env: PHPLINT=1
     - stage: deploy-to-github-dist
       env: WP_VERSION=latest
       if: tag IS present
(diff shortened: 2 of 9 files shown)


jrfnl:
Code coverage overview for the BrainMonkey based unit tests alone:
image

herregroen (approved):

PR #676, opened 2021-02-22
P1 332 genre removal alert

marijnyoast:
## Context

* Since we removed the Genres option from the News Settings, we need to notify our users about that. So an Alert was needed. For user-friendliness this Alert needed to be Dismissible. 

## Summary

This PR can be summarized in the following changelog entry:

* Adds a dismissible alert to the News Settings page to inform our users about the removal of the Genre settings.

## Relevant technical choices:

* Using the PersistentDismissableAlert container from Wordpress SEO via editorModules together with the settings store.

### Test instructions for the acceptance test before the PR gets merged
This PR can be acceptance tested by following these steps:

* Test this branch together with the P1-332 branch on Free, or for QA Free 16.0RC or higher
* Open the News Settings page and:
* * Confirm that the Alert about the removal of the Genres setting is present
* * Confirm that the Alert disappears after clicking on the dismissal button (cross)
* * Refresh the page and confirm that the Alert is still not shown

### Test instructions for QA when the code is in the RC

* [X] QA should use the same steps as above.

QA can test this PR by following these steps:

*

## Impact check

This PR affects the following parts of the plugin, which may require extra testing:

*

## UI changes

* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

## Documentation

* [ ] I have written documentation for this change.

## Quality assurance

* [X] I have tested this code to the best of my abilities
* [ ] I have added unittests to verify the code works as intended

Fixes P1-332

PR #678, opened 2021-02-25
P1-301 reactify editor

igorschoester:
## Context

* Use React for the News tab in the editor.
* Add News to the sidebar in the block editor and in Elementor.
* Merge the `Exclude` and `Google News`'s (no)index options into one (the latter).

## Summary

* Add News to the sidebar in the block editor and in Elementor.
* Changes News in the metabox.
* NON-USER-FACING Fixes an unreleased bug where a JS error would be shown on the settings pages when you have the debug flag on.

* Move hook calls from constructor to `register_hooks`, preventing immediate side effects.
* Now always registering extra meta fields. There was a post edit/new and sitemap check around this. The fields can always be registered for simplicity.
* I tried to DRY up the supported post type check, by putting it outside of the hook methods. But the post object is not available on new post pages.
* I could've split WPSEO_News_Meta_Box up in integrations: Post and Elementor split for the script loading. But the concept of integrations and loadables does not exist in News. Decided to keep it simple.
* The merging of the two options: `newssitemap-exclude` and `newssitemap-robots-index` has been done by removing the first (also from the database) and using the second where the first was used. This has been discussed and it is better this way because the `exclude` only influenced the News sitemap, and that is only for a while, since then it is not news anymore. While the `robots-index` has a more permanent effect in the meta tag being output.
  * Removing the `exclude` option from the database in a `12.7` upgrade routine.
* Added `transform-object-rest-spread` to the babel possibilities. Used in the reducer.
* JS error bugfix caused in a previous PR on this epic: https://github.com/Yoast/wpseo-news/pull/676
   * Fixed in collaboration with Marijn and improved by Josee, the width was smaller in the design
     * Tried to add `max-width` to the Alert, but the SEMrush related keyphrases modal uses the Alert slightly wider than 600px.
     * Now added inline style of `max-width: 600px` to the News settings alert wrapper. Inline to prevent having to load a separate stylesheet on this page.
   * Added commas in the echo as codescout, no string concatenation needed
* Deprecated the `add_tab_hooks` method. This was some extra methods wrapping that we don't need. We add the content with the filter in that function directly now (`yoast_free_additional_metabox_sections`).
* Add a JS window object `wpseoNewsScriptData` to use for passing data from PHP to JS in News.
* Added the 3 editors initialization in one JS file. Each have their own "check". This could be nicer in own integrations maybe. But I didn't because we don't have the PHP side split up either. If we did, it makes more sense.

#### Setup
* Important to test in the block, classic and elementor editor!
* Important to test post/page and a custom post type
* Install & activate:
  * Yoast SEO `16.0-RC4` (or newer / trunk or release branch)
    * For developer from this branch: Please link and build it with the code of this PR: https://github.com/Yoast/javascript/pull/1078 
  * Yoast SEO News `12.7-RC1` (or newer / this branch)
  * Elementor `3.1.1` (and optionally Elementor Pro `3.1.1`)
* Include post types in the News sitemap: SEO -> News SEO -> Post Types to include in News Sitemap
* Enable Elementor for those post types too: Elementor -> Settings -> Post Types
* Set your knowledge graph being represented to a company: SEO -> Search appearance -> Knowledge Graph & Schema.org and set an Organization name and Organization logo

#### Instructions to repeat per editor and post type
* Create a new post/page/CPT
* Check out the design: https://www.sketch.com/s/91b3eccb-1ec9-43c6-b705-b5bbedfdf512/a/Ze1Pav/play
  * The name of the tab no longer has the `Google` part
  * The Google news index has been visually removed
  * **Difference**: you can dismiss the information alert, which is why it has the `X`, unlike the design
  * (the genre already was removed)
* If possible (classic & block), check the metabox: there should be a tab called `News`. It should look like the design
* If possible (block & elementor), check the sidebar: there should be a collapsible called `News`. It should look like the design
* Add something in the stock ticker field
  * When in the block editor, check that when you type this, the metabox/sidebar are in sync.
* Publish the post/page/CPT and visit the frontend of it
* Check the schema output. When it contains an `Article` (not true for the Yoast Test Helper books/movies), the `@type` should have `NewsArticle` added to it
* Check the news sitemap: `http://basic.wordpress.test/news-sitemap.xml` (hint: open in new tab so you can refresh it later). There should be an entry for your post/page/CPT
* View the source of the news sitemap, you should see a `<news:stock_tickers>` entry with your content in between. Something like: `<news:stock_tickers><![CDATA[HELLO]]></news:stock_tickers>` that is part of the post/page/CPT you made
* Go back to the editor
* Change it to be excluded from Google News
* Save and visit the frontend again
* Check that:
  * Schema output no longer adds `NewsArticle` as type (if it was before)
  * There is a meta tag on the page: `<meta name="Googlebot-News" content="noindex">` (for me it was just below the schema output)
  * The post/page/CPT is no longer included in the news sitemap

#### Dismiss alert
* Go to an editor of a post/page/CPT that is News
* Go to the News tab/collapsible
* Dismiss the alert: click on the `X`
* The alert should've disappeared
* Refresh the page and verify the alert is still gone

#### Unreleased alert fix in settings
* Ensure you have `SCRIPT_DEBUG` defined in your WP configuration. Default if you are on our plugin development docker.
* Go to the News settings: SEO -> News SEO
* Do you see the alert here?
  * If not, you should reset it in your database: `wp_usermeta` -> `_yoast_alerts_dismissed` and there remove the `news-settings-genre-removal-alert` part... tricky to do correctly, or remove the whole row. (but then the one in the editor is back also).
  * Then refresh the page
* There should not be any JS warnings in the console. At least not from React.
* The alert should be 600px wide (about as wide as the publication name input).

QA can test this PR by following these steps:

*

## Impact check

*

## UI changes

* [x] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

block editor

## Documentation

## Quality assurance

Fixes https://yoast.atlassian.net/browse/P1-301

### DIFF
--- a/js/src/initializers/initializeSettingsPage.js
+++ b/js/src/initializers/initializeSettingsPage.js
@@ -10,15 +10,13 @@ const genreRemovalAlertElement = document.getElementById( "wpseo-news-genre-remo
  */
 export default function initializeSettingsPage() {
 	render(
-		<li style={{ listStyle: "none", }}>
-			<PersistentDismissableAlert
-				store="yoast-seo/settings"
-				type="info"
-				alertKey="news-settings-genre-removal-alert"
-			>
-				{ __( "Google no longer supports 'Genres' for articles in Google News, therefore we decided to remove the 'Default genre' setting below.", "wordpress-seo-news" ) }
-			</PersistentDismissableAlert>
-		</li>,
-		genreRemovalAlertElement
+		<PersistentDismissableAlert
+			store="yoast-seo/settings"
+			type="info"
+			alertKey="news-settings-genre-removal-alert"
+		>
+			{ __( "Google no longer supports 'Genres' for articles in Google News, therefore we decided to remove the 'Default genre' setting below.", "wordpress-seo-news" ) }
+		</PersistentDismissableAlert>,
+		genreRemovalAlertElement,
 	);
 }
--- a/js/src/initializers/initializeSettingsPage.js
+++ b/js/src/initializers/initializeSettingsPage.js
@@ -1,23 +1,24 @@
 import { render } from "@wordpress/element";
 import { __ } from "@wordpress/i18n";
+
 const PersistentDismissableAlert = window.yoast.editorModules.containers.PersistentDismissableAlert;
 
 const genreRemovalAlertElement = document.getElementById( "wpseo-news-genre-removal-alert" );
 
 /**
  * Renders the React content on the Settings pages.
- *
- * @returns {Object} React Element.
  */
 export default function initializeSettingsPage() {
 	render(
-		<PersistentDismissableAlert
-			store="yoast-seo/settings"
-			type="info"
-			alertKey="news-settings-genre-removal-alert"
-		>
-			{ __( "Google no longer supports 'Genres' for articles in Google News, therefore we decided to remove the 'Default genre' setting below.", "wordpress-seo-news" ) }
-		</PersistentDismissableAlert>,
+		<li style={{ listStyle: "none", }}>
+			<PersistentDismissableAlert
+				store="yoast-seo/settings"
+				type="info"
+				alertKey="news-settings-genre-removal-alert"
+			>
+				{ __( "Google no longer supports 'Genres' for articles in Google News, therefore we decided to remove the 'Default genre' setting below.", "wordpress-seo-news" ) }
+			</PersistentDismissableAlert>
+		</li>,
 		genreRemovalAlertElement
 	);
 }
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,8 @@
       {
         "pragma": "wp.element.createElement"
       }
-    ]
+    ],
+    "transform-object-rest-spread"
   ],
   "env": {
     "test": {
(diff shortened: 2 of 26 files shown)


js/src/initializers/initializeSettingsPage.js
```diff
@@ -1,23 +1,24 @@
...
 export default function initializeSettingsPage() {
 	render(
-		<PersistentDismissableAlert
-			store="yoast-seo/settings"
-			type="info"
-			alertKey="news-settings-genre-removal-alert"
-		>
-			{ __( "Google no longer supports 'Genres' for articles in Google News, therefore we decided to remove the 'Default genre' setting below.", "wordpress-seo-news" ) }
-		</PersistentDismissableAlert>,
+		<li style={{ listStyle: "none", }}>
```
JoseeWouters, line 13: Please remove this `li` here.

JoseeWouters (changes requested):
Some minor improvements needed

js/src/initializers/initializeSettingsPage.js
igorschoester: As discussed, now replaced by `max-width: 600px` on the `Alert`.

JoseeWouters (approved):
CR 👍

marijnyoast:
Acceptance 🥇

PR #679, opened 2021-02-25
Add the ability to filter News Sitemap items for exclusion

joneslloyd:
## Context
We have a requirement to skip all posts that _don't_ have a specific taxonomy term( "News"). Adding this filter would give us the ability to add our required functionality, and we'd therefore be able to purchase this add-on (otherwise we have to implement the same functionality in-house).

## Summary
changelog: enhancement

N/A(?)

## Test instructions
Use the `Yoast\WP\News\skip_build_item` filter to exclude a CPT that has passed the other requirements for not being excluded. The two params are the current exclude status (`false`), and the item ID.

N/A

N/A

## Impact check
N/A

## UI changes

## Documentation

## Quality assurance

Fixes #

### DIFF
--- a/classes/sitemap-item.php
+++ b/classes/sitemap-item.php
@@ -89,7 +89,7 @@ class WPSEO_News_Sitemap_Item {
 			return true;
 		}
 
-		return apply_filters( 'Yoast\WP\News\skip_build_item', $this->item->ID );
+		return apply_filters( 'Yoast\WP\News\skip_build_item', false, $this->item->ID );
 	}
 
 	/**
--- a/classes/sitemap-item.php
+++ b/classes/sitemap-item.php
@@ -89,7 +89,7 @@ class WPSEO_News_Sitemap_Item {
 			return true;
 		}
 
-		return false;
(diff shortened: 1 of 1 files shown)


classes/sitemap-item.php
```diff
@@ -89,7 +89,7 @@ private function skip_build_item() {
 			return true;
 		}
 
-		return false;
+		return apply_filters( 'Yoast\WP\News\skip_build_item', false, $this->item->ID );
```
jdevalk, line 92: Why would we only filter `false` here, and not `true` `return`s?

joneslloyd: Good question. I only patched this for my narrow use-case. Definitely not an optimal solution.

maartenleenders:
Thanks for the assist, @joneslloyd ! 

I've added the filter `'Yoast\WP\News\skip_build_item'` as you suggested, with minor modifications to also filter the true values. If all goes well, it should be in wpseo-news 12.8 😄, hope that helps.

joneslloyd:
No worries :) Glad I could be of help

PR #680, opened 2021-03-05
P3-362 Move reading of options outside functions called by options cache priming, to avoid loops

### LINKED ISSUE #660: Yoast SEO News introduces infinite recursion while Yoast SEO rebuilds its option cache
### What did you expect to happen?

The site not to crash. 

### What happened instead?

PHP-FPM process crashes with segmentation fault.

Install wordpress, Yoast SEO and Yoast SEO News plugins. Install and enable php xdebug module. No additional configuration is needed for xdebug. Set WP_DEBUG to true, so you can see xdebug errors displayed on page. 

**Important part:** Your user locale should be different than the Site language locate in site `Settings -> General -> Site Language` section. For example if the site is set in `English`, the user can be set to use `German` locale. Go to `WP Admin -> Appearance -> Customize` to preview the theme. You will get xdebug error `Maximum function nesting level of '256' reached`.  

256 is the default value for `xdebug.max_nesting_level` xdebug setting. You can increase the number, but it is not gonig to help. If xdebug module is not enabled, there will be no error when the page loads . It is possible the PHP-FPM process to crash with segmentation fault and get 500 status code. This is when I started the investigation. I didn't get segmentation fault on my local VM. It is possible PHP-FPM crash to be because of another loaded PHP module.

You can see the loop in the xdebug stacktrace, but to save you time if you comment `add_filter( 'wpseo_schema_article_post_types', [ $this, 'article_post_types' ] );` in `schema.php` file the loop will disappear. The reason is `article_post_types` is executed when `wpseo_titles` is read to be put in the cache. `article_post_types` indirectly calls a function `is_excluded_through_terms` in `wpseo-news.php` file that cas a call to `WPSEO_Options::get( 'news_sitemap_exclude_terms', [] )`. But we are already rebuilding the cache and the loop starts.

This only happens when the Yoast option cache is empty. Having the wp user a different locale than the site, `WP_Customize_Manager` class triggers `change_locale` action. `change_locale` action triggers `create_initial_post_types` which on triggers purging Yoast SEO option cache. If the use locale is the same as the site locale, the loop doesn't happen.

Going to `Customize` is one example that I am aware of to hit that edge case with infinite loop. There might be other cases.

### Technical info
* WordPress version: 5.5.3
* Yoast SEO version: 15.3
* Yoast SEO News version: 12.6

enricobattocchi:
## Context

* See #660 for description of the issue

## Summary

* Fixes a bug where certain conditions (e.g. using a different admin language) would result in an endless loop.

* The loop was caused by the querying of the options cache in functions called hooked to the WP `get_options`. The variables storying the options value were converted to static properties since they are accessed by static methods. This is not optimal, also for testing purposes.

* in `Settings` > `General`, install a different language from the default one. Then, switch back to the default language, but this way you will have the chance to select a different language for your user.
* In `Users` > `Profile`, change the language of your account and save
* visit `Appearance` > `Customize` and observe that everythings looks good and you don't get any fatal error for Maximum nested functions or a `Bad gateway` page.

QA can test this PR by following these steps:

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes #660, [P3-362]

### DIFF
--- a/classes/wpseo-news.php
+++ b/classes/wpseo-news.php
@@ -78,7 +78,7 @@ class WPSEO_News {
 	/**
 	 * Populates static properties from options so they have to be queries each time we need them.
 	 */
-	public static function read_options(){
+	public static function read_options() {
 		self::$included_post_types = (array) WPSEO_Options::get( 'news_sitemap_include_post_types', [] );
 		self::$excluded_terms      = (array) WPSEO_Options::get( 'news_sitemap_exclude_terms', [] );
 	}
--- a/classes/wpseo-news.php
+++ b/classes/wpseo-news.php
@@ -20,6 +20,20 @@ class WPSEO_News {
 	const VERSION = WPSEO_NEWS_VERSION;
 
 	/**
+	 * Included post types.
+	 *
+	 * @var array
+	 */
+	public static $included_post_types = [];
+
+	/**
+	 * Excluded terms.
+	 *
+	 * @var array
+	 */
+	public static $excluded_terms = [];
+
+	/**
 	 * Initializes the plugin.
 	 */
 	public function __construct() {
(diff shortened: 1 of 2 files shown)


classes/wpseo-news.php
```diff
@@ -19,6 +19,20 @@ class WPSEO_News {
 	 */
 	const VERSION = WPSEO_NEWS_VERSION;
 
+	/**
+	 * Included post types.
+	 *
+	 * @var array
+	 */
+	public static $included_post_types = [];
```
d-claassen, line 27: Can we keep these new constants (`$included_post_types` and `$excluded_terms`) protected to keep them out of the public api for this class?

classes/wpseo-news.php
```diff
@@ -60,6 +75,14 @@ private function set_hooks() {
 		add_filter( 'wpseo_frontend_presenters', [ $this, 'add_frontend_presenter' ] );
 	}
 
+	/**
+	 * Populates static properties from options so they have to be queries each time we need them.
```
d-claassen, line 79:
```suggestion
	 * Populates static properties from options so they don't have to be queried each time we need them.
```

d-claassen (changes requested):

classes/wpseo-news.php
enricobattocchi: Sure, will fix!

classes/wpseo-news.php
enricobattocchi: Yes, definitely. Thanks!

d-claassen (approved):
Awesome. CR done! ✅

jcomack:
Acceptance done 👍

PR #691, opened 2021-05-05
Adjust sitemap query to improve performance

### LINKED ISSUE #690: Extremely inefficient but easily fixable query in News SEO
There's a query in WP SEO here https://github.com/Yoast/wpseo-news/blob/trunk/classes/sitemap.php#L233 that looks like this:

```
		// Get posts for the last two days only, credit to Alex Moss for this code.
		$items = $wpdb->get_results(
			$wpdb->prepare(
				"SELECT ID, post_content, post_name, post_author, post_parent, post_date_gmt, post_date, post_date_gmt, post_title, post_type
				FROM {$wpdb->posts}
				WHERE post_status='publish'
					AND ( TIMESTAMPDIFF( MINUTE, post_date_gmt, UTC_TIMESTAMP() ) <= ( 48 * 60 ) )
					AND post_type IN (" . implode( ',', array_fill( 0, count( $post_types ), '%s' ) ) . ')
				ORDER BY post_date_gmt DESC
				LIMIT 0, %d
				',
				$replacements
			)
		);
```

The problem with this query is it is extremely slow due to `TIMESTAMPDIFF` and `UTC_TIMESTAMP` calculations. I'm talking a full minute for a table of 700k wp_posts rows. It's probably a lot worse with busier dbs.

But this can be easily fixed with a slight tweak - and this query runs in less than 0.1s. Replace `TIMESTAMPDIFF( MINUTE, post_date_gmt, UTC_TIMESTAMP() ) <= ( 48 * 60 )` with `post_date_gmt >= NOW() - INTERVAL 48 HOUR` and voila. Same results, but at 1000x the speed.

Please consider pushing this change out ASAP after your tests.

Thank you.

herregroen:
## Context

* Improve the sitemap query performance.

## Summary

* Improves the performance of the news sitemap. Props to [archon810](https://github.com/archon810).

* Make sure you have a large selection of posts that have a valid post type for the news sitemap ( these are set in the settings ).
* This should include posts more than 2 days old ( these should not show in the sitemap ).
* The output of the news sitemap should be identical, with all valid items less than 2 days old appearing and valid items more than 2 days old not appearing.

## Impact check

* The news sitemap.

## UI changes

## Documentation

## Quality assurance

Fixes #690

### DIFF
--- a/classes/sitemap.php
+++ b/classes/sitemap.php
@@ -230,7 +230,7 @@ class WPSEO_News_Sitemap {
 				"SELECT ID, post_content, post_name, post_author, post_parent, post_date_gmt, post_date, post_date_gmt, post_title, post_type
 				FROM {$wpdb->posts}
 				WHERE post_status='publish'
-					AND ( TIMESTAMPDIFF( MINUTE, post_date_gmt, UTC_TIMESTAMP() ) <= ( 48 * 60 ) )
+					AND post_date_gmt >= NOW() - INTERVAL 48 HOUR
 					AND post_type IN (" . implode( ',', array_fill( 0, count( $post_types ), '%s' ) ) . ')
 				ORDER BY post_date_gmt DESC
 				LIMIT 0, %d
--- a/integration-tests/sitemap-test.php
+++ b/integration-tests/sitemap-test.php
@@ -238,7 +238,7 @@ class WPSEO_News_Sitemap_Test extends WPSEO_News_UnitTestCase {
 			]
 		);
 
-		$two_days_ago = strtotime( '-48 hours' );
+		$two_days_ago = strtotime( '-47 hours -59 minutes' );
 
 		$this->factory->post->create(
 			[


johannadevos:
CR: ✅ 

Acceptance: ✅ 
- [x] Between `trunk` and this branch, the generated sitemap is the same (and only includes posts from the last 2 days).
- [ ] Ideally we would also test the performance impact of this PR, since that is what it is actually about.

The integration test `test_sitemap_only_showing_recent_items()` still needs fixing. I found a potential solution by changing line 241 to `$two_days_ago = strtotime( '-48 hours +1 minute' );`, but I'm not sure if this is acceptable. Will be looked into later.

johannadevos:
See https://developer.yoast.com/development/environment/setup-plugin-integration-tests for setting up the integration tests

hansjovis:
Acceptance test: ✅

PR #703, opened 2021-07-01
P2 1093 implement abstract indexable presenter

increddibelly:
## Context

* Made News compatible with new Abstract Indexable Tag Presenter in free 16.7

## Summary

* Adds News SEO data to the new json head fields.

* removed a deprecated filter
* moved present logic to get method
* added key and tag format to make the presenter compatible with the `Abstract_Indexable_Tag_Presenter` that's new in free 16.7

by default, this PR doesn't do anything, because empty values are not output in the JSON head fields.
* access a post via the rest API; 
* visit https://basic.wordpress.test/wp-json/wp/v2/posts
   output starts with [{"id":xxxxx -> use xxxxx as the Id for the next steps.
* visit http://[your-wordpress-server]/wp-json/wp/v2/posts/xxxxx
* copy & paste the entire output in a json prettifier e.g. https://jsonformatter.curiousconcept.com/
* the json output should show a yoast_head_json property, with no `Googlebot-News` key 

To make the the Googlebot-News-Presenter output anything:
this can only be done by third party integrations using filters. we'll do it by hand here: 
* open the `classes\googlebot-news-presenter.php` file
* change the first line of the `public function get()` function to `return 'noindex';` note that this gives the exact same output as when the filter would return a positive result.
* repeat the above rest api call, visit http://[your-wordpress-server]/wp-json/wp/v2/posts/xxxxx
* note the yoast_head_json field now contains a `Googlebot_News` key with the text value `noindex`; 

* [ ] QA should use the same steps as above.

QA have to also test the additional item below when testing the RC:

* Confirm that the Yoast Free plugin is NOT added in the  `/vendor` dir in the RC (much like it doesn't, with already released versions).

## Impact check

* headless rest api

## UI changes

## Documentation

## Quality assurance

Fixes #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ jobs:
         - chmod 600 ./deploy_keys/wpseo_news_deploy
         - eval $(ssh-agent -s)
         - ssh-add ./deploy_keys/wpseo_news_deploy
+        - composer config --global github-oauth.github.com $GITHUB_OAUTH_TOKEN
       before_deploy:
         - nvm install node
         - curl -o- -L https://yarnpkg.com/install.sh | bash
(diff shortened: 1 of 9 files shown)


hansjovis (approved):
CR: 👍 

It misses some test instructions, though.

leonidasmi (approved):
Acceptance: 👍

PR #721, opened 2021-10-22
Move to an Indexables based XML News Sitemap

jdevalk:
Requires https://github.com/Yoast/wordpress-seo/pull/17563

## Summary

* Huge performance improvement: move the XML News sitemap to be based on our Indexables architecture. 
* Remove images from the XML News sitemap as they serve no purpose here and this further improves performance.

* XML News sitemap output should still contain posts from the last 48 hours, dependent on all settings (noindex or not, removed from XML news sitemap or not, etc).
* Excluded terms are changed in the upgrade routine, you should set excluded terms, upgrade to this PR and then validate your settings are maintained.
* All indexables from the last 48 hours should have their object_published_at and object_last_modified set after running the upgrade routine.

## Impact check

* This is a big one. Please test it very thoroughly.

## Quality assurance

Fixes P2-1270
Fixes P2-1271
Fixes P2-1272
Fixes P2-1240


enricobattocchi:
The integration test for the sitemap item has been removed and will be reinstated as a unit test in a next issue.

